PR #40, opened 2022-03-08
Adding registry sync and tests

jainr:


feathr_project/feathr/_feature_registry.py
```diff
@@ -578,3 +579,42 @@ def list_registered_features(self, project_name: str = None) -> List[str]:
                 feature_list.append(entity["name"])
 
         return feature_list
+    
+    def sync_features_from_registry(self, project_name: str, workspace_path: str):
```
xiaoyongzhu, line 583: Should we say something like "get_features_from_registry" so that it's clear on the direction?

feathr_project/feathr/client.py
```diff
@@ -442,3 +442,7 @@ def _get_sql_config_str(self):
             JDBC_PASSWORD: "{JDBC_PASSWORD}"
             """.format(JDBC_TABLE=table, JDBC_USER=user, JDBC_PASSWORD=password)
         return config_str
+
+    def sync_features(self, project_name):
```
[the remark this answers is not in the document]
jainr: Good point, addressed.

feathr_project/feathr/_feature_registry.py
jainr: Good point, addressed.

xiaoyongzhu (dismissed):

feathr_project/feathr/client.py
```diff
@@ -442,3 +442,7 @@ def _get_sql_config_str(self):
             JDBC_PASSWORD: "{JDBC_PASSWORD}"
             """.format(JDBC_TABLE=table, JDBC_USER=user, JDBC_PASSWORD=password)
         return config_str
+
+    def get_features_from_registry(self, project_name):
+        """ Sync features from the registry given a project name """      
+        self.registry.get_features_from_registry(project_name, os.path.abspath("./"))
```
Yuqing-cat, line 448: Is this by design that specify workspace path is not exposed to user? They can only achieve so, by execute code under certain workspace folder.

Yuqing-cat (approved):
Ability to customized workspace path may be an enhancement to the whole python client.

feathr_project/feathr/client.py
jainr: Added TODO to address this

xiaoyongzhu (approved):

Yuqing-cat (approved):

PR #53, opened 2022-03-14
Add python api

jaymo001:


feathr_project/feathr/client.py
```diff
@@ -251,7 +256,27 @@ def _construct_redis_client(self):
...
+        tm = Template("""
+            {{observation_settings.to_config()}}
+            featureList: [
+                {% for list in feature_lists %}
+                    {{list.to_config()}}
+                {% endfor %}
+            ]
+        """)
+        feature_queries = feature_query if isinstance(feature_query, List) else [feature_query]
+        config = tm.render(feature_lists=feature_queries, observation_settings=observation_settings)
```
xiaoyongzhu, line 273: I assume even if I don't register the features, the offline features will save all the feature objects into templates and submit the spark job?

feathr_project/feathr/_feature_registry.py
```diff
@@ -530,12 +545,84 @@ def _read_config_from_workspace(self, workspace_path: str):
 
         self.entity_batch_queue.append(workspace)
 
-    def register_features(self, workspace_path: str = None):
+    def get_py_files(self, path: Path) -> List[Path]:
```
hangfei, line 548: get_py_feature_files?
does this need to be public?

feathr_project/feathr/typed_key.py
```diff
@@ -0,0 +1,54 @@
...
+        return result
+
+    def __deepcopy__(self, memo):
+        cls = self.__class__
+        result = cls.__new__(cls)
+        memo[id(self)] = result
+        for k, v in self.__dict__.items():
+            setattr(result, k, deepcopy(v, memo))
+        return result
+
```
hangfei, line 44: could you doc why we need copy and deep copy for future devs.

feathr_project/feathr/job_utils.py
```diff
@@ -0,0 +1,19 @@
+from feathr.client import FeathrClient
+import os
+import glob
+import pandavro as pdx
+import pandas as pd
+import tempfile
+
+
+def get_result_df(client: FeathrClient) -> pd.DataFrame:
```
hangfei, line 9: get_result_dataframe

hangfei: add documentation.

feathr_project/feathr/feature_derivations.py
```diff
@@ -0,0 +1,93 @@
...
+        result.__dict__.update(self.__dict__)
+        return result
+
+    def __deepcopy__(self, memo):
+        cls = self.__class__
+        result = cls.__new__(cls)
+        memo[id(self)] = result
+        for k, v in self.__dict__.items():
+            setattr(result, k, deepcopy(v, memo))
+        return result
```
hangfei, line 74: can this be a shared util with TypedKey?

feathr_project/feathr/feature_derivations.py
```diff
@@ -0,0 +1,93 @@
...
+
+from copy import copy, deepcopy
+from jinja2 import Template
+from typing import Union
+from feathr.dtype import FeatureType, ValueType
+from feathr.feature import Feature
+from feathr.typed_key import DUMMY_KEY, TypedKey
+from feathr.transformation import ExpressionTransformation, RowTransformation
+
+class DerivedFeature:
```
hangfei, line 12: Feature and Derived feature should have a base class that have: feature type, feature name, feature description.

feathr_project/feathr/feature.py
```diff
@@ -1,27 +1,87 @@
...
+        transform: A row transformation used to produce its feature value. e.g. amount * 10
+        feature_alias: variable name, transformation can reference this feature by the feature_alias. Default to its feature name.
+        entity_alias: An alias for the entity in this feature. Default to its entity alias. Useful in complex relation features.
+    """
     def __init__(self,
                 name: str,
                 feature_type: FeatureType,
-                transform: Optional[RowTransformation] = None):
+                key: Optional[Union[TypedKey, List[TypedKey]]] = [DUMMY_KEY],
+                transform: Optional[Union[str, RowTransformation]] = None,
```
hangfei, line 29: what about WindowAggTransformation?

feathr_project/feathr/client.py
jaymo001: Updated

feathr_project/feathr/anchor.py
```diff
@@ -33,16 +39,9 @@ def to_feature_config(self) -> str:
...
-        
-        {% if not source.name == "PASSTHROUGH" %}
-        sources: {
-            {{source.to_feature_config()}}
-        }
-        {% endif %}
-        """)
-        key_list = ','.join(key for key in self.key_columns)
+        """)     
+        key_list = ','.join(key for key in self.features[0].key_alias)
```
LeBenHL, line 43: I think we should keep the key provided at the `FeatureAnchor` level instead. It makes it explicit that all Features in the FeatureAnchor shares the same key

Otherwise, I think we should verify that the key is consistent across all features put in the same anchor and error otherwise.

I'm ok with both

feathr_project/feathr/feature_derivations.py
LeBenHL: +1

feathr_project/feathr/settings.py
```diff
@@ -0,0 +1,33 @@
+from typing import Optional
+from jinja2 import Template
+
+
+class ObservationSettings:
+    """Time settings of the observation data. Used in feature join."""
```
LeBenHL, line 6: Missing pydoc. What is an `event_timestamp_column`? We need to explain this to users. Please audit and make sure documentation standards are met on all user facing Python Config APIs

feathr_project/feathr/typed_key.py
```diff
@@ -0,0 +1,54 @@
+from __future__ import annotations
+from copy import copy, deepcopy
+
+from typing import List, Optional
+from feathr.dtype import ValueType, FeatureType
+
+
+class TypedKey:
+    """A feature is an individual measurable property or characteristic of an 'Entity'.
```
LeBenHL, line 9: This is not an Entity anymore, please don't introduce new concepts yet so we don't have to rush to design and implement them

feathr_project/feathr/typed_key.py
```diff
@@ -0,0 +1,54 @@
...
+      e.g. the price is a feature of a car, and the car is an 'entity'.
+      Attributes:
+        key_column: The id column name(s) of this entity. e.g. 'product_id'. The value(s)
+                    of the id column(s) should identify a unique instance of the entity.
+        key_column_type: Types of the key_column
+        full_name: Unique name of the entity. Recommend using [project_name].[entity_name], e.g. foo.bar
+        description: Documentation for the entity.
+        key_column_alias: Used in relation feature. Default to the full_name with its '.' replaced with '_'.
```
LeBenHL, line 17: Outdated, default is the `key_column`

feathr_project/feathr/feature.py
```diff
@@ -1,27 +1,87 @@
...
+
+    def __deepcopy__(self, memo):
+        cls = self.__class__
+        result = cls.__new__(cls)
+        memo[id(self)] = result
+        for k, v in self.__dict__.items():
+            setattr(result, k, deepcopy(v, memo))
+        return result
+
+    def with_key(self, key_alias: Union[str, List[str]]) -> Feature:
```
LeBenHL, line 60: Should with_key take a TypedKey object as input instead of a string? Seems like that is more sensible

feathr_project/feathr/feature_derivations.py
```diff
@@ -0,0 +1,93 @@
...
+                feature_type: FeatureType,
+                input_features: Union[DerivedFeature, Feature, List[Union[DerivedFeature, Feature]]],
+                transform: Union[str, RowTransformation],
+                key: Optional[Union[TypedKey, List[TypedKey]]] = [DUMMY_KEY],
+                feature_alias:Optional[str] = None,
+                key_alias: Optional[List[str]] = None) -> None:
+        self.name = name
+        self.feature_type = feature_type
+        self.key = key if isinstance(key, List) else [key]
+        self.input_features = input_features if isinstance(input_features, List) else [input_features]
```
LeBenHL, line 35: Can we do some validation such as that the keys used by the `input_features` are in the `key` list/set?

This way, we can prevent user from shooting themselves in the foot

feathr_project/feathr/feature_derivations.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+
+    def __init__(self,
+                name: str,
+                feature_type: FeatureType,
+                input_features: Union[DerivedFeature, Feature, List[Union[DerivedFeature, Feature]]],
+                transform: Union[str, RowTransformation],
+                key: Optional[Union[TypedKey, List[TypedKey]]] = [DUMMY_KEY],
+                feature_alias:Optional[str] = None,
+                key_alias: Optional[List[str]] = None) -> None:
```
LeBenHL, line 31: Should `key_alias` be exposed as a part of the initializer? Seems like it should only be edited by the `as_feature` method

feathr_project/feathr/settings.py
```diff
@@ -0,0 +1,39 @@
...
+    """Time settings of the observation data. Used in feature join.
+    Attributes:
+        observation_path: path to the observation dataset, i.e. input dataset to get with features
+        output_path: output path of job, i.e. the observation data with features attached.
+        event_timestamp_column: column name of the event timestamp
+        timestamp_format: the format of the event_timestamp_column, e.g. yyyy/MM/DD.
+    """
+    def __init__(self,
+                 observation_path: str,
+                 output_path: str,
```
hangfei, line 15: output path should not be part of the observationSettings?

It's hard to explain in documentation as well.

feathr_project/feathr/feature.py
jaymo001: It should be strings, as they are just 'alias', not actually new TypedKeys.

feathr_project/feathr/client.py
```diff
@@ -251,7 +256,27 @@ def _construct_redis_client(self):
...
+        config_file_name = "feature_join_conf/feature_join.conf"
```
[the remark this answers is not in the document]
jaymo001: Sync offline, no need to change.

xiaoyongzhu (dismissed):

hangfei (dismissed):

hangfei (approved):

xiaoyongzhu (dismissed):

xiaoyongzhu (approved):

feathr_project/test/test_derived_features.py
```diff
@@ -0,0 +1,187 @@
...
+    }"""
+    assert_config_equals(user_item_similarity.to_feature_config(), derived_feature_config)
+
+
+def test_derived_feature_to_config_with_alias():
+    # More complicated use case, viewer viewee aliasged user key
+    # References the same key feature with different alias
+    user_key = TypedKey(full_name="mockdata.user", key_column="user_id", key_column_type=ValueType.INT32, description="An user identifier")
+    user_embedding = Feature(name="user_embedding", key=user_key, feature_type=FLOAT_VECTOR)
+    viewer_viewee_distance = DerivedFeature(name="viewer_viewee_distance",
```
LeBenHL, line 77: I was thinking something like
```
class FeatureQuery:
  
    def __init__(self, feature: Feature, key: Optional[Union[TypedKey, List[TypedKey]]] = None, alias: Optional[str] = None) -> None:
        self.feature = feature
        self.key = key
        self.alias = alias

viewer_key = TypedKey(..., key_column="viewer_id")
viewee_key = TypedKey(..., key_column="viewee_id")
viewer_viewee_distance = DerivedFeature(name="viewer_viewee_distance",
                    key=[viewer_key, viewee_key],
                    feature_type=FLOAT,
                    input_features=[FeatureQuery(feature=user_embedding, key=viewer_key, alias="viewer_embedding"),
                                    FeatureQuery(feature=user_embedding, key=viewee_key, alias="viewee_embedding")],
                    transform="distance(viewer_embedding, viewee_embedding)")
```

FeatureQuery is query on a per feature level
You can use FeatureQuery as an input to a derived feature
The key tag notation in derived features is trying to express a FeatureQuery essentially?
Makes it also explicit that the same key object is used for the DerivedFeature key, as well as for the query for the input feature. I think it also reflects more clearly on what the HOCON config version of key-tagged features are
```
      inputs: {
            viewer_embedding: {key: [viewer], feature: user_embedding}
            viewee_embedding: {key: [viewee], feature: user_embedding}
        }
```
This is a feature query, using a key query (e.g. viewer or viewee) to retrieve some feature by name (e.g. user_embedding) and doing an optional renaming .

feathr_project/feathrcli/data/feathr_user_workspace/features/request_features.py
```diff
@@ -0,0 +1,36 @@
...
+    f_trip_time_duration,
+    Feature(name="f_is_long_trip_distance",
+            feature_type=BOOLEAN,
+            transform="cast_float(trip_distance)>30"),
+    Feature(name="f_day_of_week",
+            feature_type=INT32,
+            transform="dayofweek(lpep_dropoff_datetime)"),
+  ]
+
+request_anchor = FeatureAnchor(name="request_features",
```
d4ve, line 23: I’m unsure about the idiom we are using here. Wouldn’t the feature definition script look more intuitive if each “passthrough” `Feature` here was self-contained? User may wonder why they need to invoke `FeatureAnchor` here, and why they can’t just specify the `Feature`s and be done with it.
Something that looks “off” about this IMO is that the expressions like `transform=“trip_distance”` are isolated from the context about what `source` is actually being used, even though they can’t really be understood without that context.
One possible “fix” might just be via style recommendation, e.g. we can encourage users to invoke all those `Feature` constructors etc. inline here within the `FeatureAnchor` constructor invocation (which incidentally would look more like Frame HOCON). If we want to do that, we should adopt this style in examples like this, and in the docs.

LeBenHL: Maybe a crazy thought is to just attach source to the feature level? And hide the FeatureAnchor concept completely?

When building the config file in `RepoDefinitions`, we can then just auto group features in the same source together and synthesize the feature anchor there?

I know this is slightly different from the HOCON concept where we do feature group authoring as the level of atomicity but I think this makes sense to do so as a declarative API to attach the same source to multiple features and "optimize" it by grouping common source things together as part of the config compilation step.

d4ve: +1 to Ben's idea about attaching it to the feature level. @jaymo001 WDYT?

xiaoyongzhu:
Close as it was merged using another PR

PR #65, opened 2022-03-16
Improving CI pipeline

xiaoyongzhu:


Yuqing-cat (approved):
May be we can split them into 2 parallel jobs in the future to save test time

hangfei (approved):

PR #90, opened 2022-03-24
Enable set up spark configurations in `FeathrClient` and enable multiple read/write format

xiaoyongzhu:
In this PR, the goal is to:
- Expose the APIs so end users can directly setup spark configurations, in order to allow customized behaviors
- enable multiple read/write format by setting `spark.feathr.outputFormat` and `spark.feathr.inputFormat`, to allow input/output format such as parquet/json/etc.


src/main/scala/com/linkedin/feathr/offline/job/FeatureJoinJob.scala
```diff
@@ -157,7 +157,12 @@ object FeatureJoinJob {
     val (joinedDF, _) = getFeathrClientAndJoinFeatures(ss, observationsDF, featureGroupings, joinConfig, jobContext, localTestConfig)
 
 
-    val parameters = Map(SparkIOUtils.OUTPUT_PARALLELISM -> jobContext.numParts.toString, SparkIOUtils.OVERWRITE_MODE -> "ALL")
+    var parameters = Map(SparkIOUtils.OUTPUT_PARALLELISM -> jobContext.numParts.toString, SparkIOUtils.OVERWRITE_MODE -> "ALL")
+
+    if (ss.conf.get("spark.feathr.outputFormat", "").nonEmpty) {
```
hangfei, line 162: create a val for ss.conf.get("spark.feathr.outputFormat", "")

src/main/scala/com/linkedin/feathr/offline/source/dataloader/hdfs/FileFormat.scala
```diff
@@ -71,7 +71,11 @@ object FileFormat {
         ss.read.format(ORC_DATASOURCE).load(existingHdfsPaths: _*)
       case PARQUET =>
         ss.read.parquet(existingHdfsPaths: _*)
-      case _ => throw new FeathrException(s"Unsupported data format $format. Only AVRO and ORC are supported.")
+      case _ =>
+        // if we cannot detect the file type by extension, we will detect "spark.feathr.inputFormat" and use that as the option; this is a global config (i.e. affecting all the files) so customers should use it as the last resort.
+        // If this is not set, throw an exception
+        if (ss.conf.get("spark.feathr.inputFormat", "").nonEmpty) ss.read.format(ss.conf.get("spark.feathr.inputFormat")).load(existingHdfsPaths: _*)
```
hangfei, line 77: the naming is not consistent with output.format

src/main/scala/com/linkedin/feathr/offline/job/FeatureJoinJob.scala
```diff
@@ -157,7 +157,12 @@ object FeatureJoinJob {
...
+      parameters = parameters + (SparkIOUtils.OUTPUT_FORMAT -> ss.conf.get("spark.feathr.outputFormat"))
```
hangfei, line 163: why not just output.format?

src/main/scala/com/linkedin/feathr/offline/source/dataloader/hdfs/FileFormat.scala
hangfei, line 77: this is a bit magic. When will we trigger this code path?

hangfei (commented):
Could you describe what tests have you done in the PR msg.

Also could we add a test for the code you have changed? Thanks.

src/main/scala/com/linkedin/feathr/offline/job/FeatureJoinJob.scala
xiaoyongzhu: guess I'm trying to be consistent with other code

src/main/scala/com/linkedin/feathr/offline/source/dataloader/hdfs/FileFormat.scala
xiaoyongzhu: It should be consistent, i.e. spark.feathr.inputFormat and spark.feathr.outputFormat

xiaoyongzhu: This is like the last resort, i.e. if all other options cannot parse user's input format, we will use the value of spark.feathr.inputFormat to read the data (for example if users use `json`)

xiaoyongzhu:
> Could you describe what tests have you done in the PR msg.
> 
> Also could we add a test for the code you have changed? Thanks.

Test added and comments resolved.

feathr_project/feathr/client.py
```diff
@@ -349,14 +349,16 @@ def _construct_redis_client(self):
     def get_offline_features(self,
                              observation_settings: ObservationSettings,
                              feature_query: Union[FeatureQuery, List[FeatureQuery]],
-                             output_path: str
+                             output_path: str,
+                             execution_configuratons: Dict[str,str] = None,
```
jaymo001, line 353: Can we create a SparkExecutionConfiguration which wraps Dict[str, str] ?

src/main/scala/com/linkedin/feathr/offline/source/dataloader/hdfs/FileFormat.scala
```diff
@@ -58,7 +60,10 @@ object FileFormat {
   // existingHdfsPaths may be folder or file with suffix
   // Currently only support parquet file but not folder with parquet files
   def getTypeForUnionDF(existingHdfsPaths: Seq[String], dataIOParameters: Map[String, String] = Map()): String = {
-    if (existingHdfsPaths.head.endsWith(".parquet")) PARQUET else dataIOParameters.getOrElse(DATA_FORMAT, AVRO).toUpperCase
+    // if we cannot detect the file type by extension, we will detect "spark.feathr.inputFormat" and use that as the option; this is a global config (i.e. affecting all the files) so customers should use it as the last resort.
+    // If this is not set, throw an exception
+    if (existingHdfsPaths.head.endsWith(".parquet")) PARQUET else dataIOParameters.getOrElse(DATA_FORMAT, ss.conf.get("spark.feathr.inputFormat", AVRO)).toUpperCase
+//    if (existingHdfsPaths.head.endsWith(".parquet")) PARQUET else dataIOParameters.getOrElse(DATA_FORMAT, AVRO).toUpperCase
```
jaymo001, line 66: Delete this?

feathr_project/feathr/client.py
```diff
@@ -399,12 +401,16 @@ def _get_offline_features_with_config(self, feature_join_conf_path='feature_join
...
+            job_tags[OUTPUT_FORMAT]= execution_configuratons[OUTPUT_FORMAT]
 
         # submit the jars
         return self.feathr_spark_laucher.submit_feathr_job(
             job_name=self.project_name + '_feathr_feature_join_job',
             main_jar_path=self._FEATHR_JOB_JAR_PATH,
-            job_tags={OUTPUT_PATH_TAG:feature_join_job_params.job_output_path},
+            job_tags=job_tags,
             main_class_name='com.linkedin.feathr.offline.job.FeatureJoinJob',
             arguments=[
```
hangfei, line 415: why are we using tgs and some other times using arguments to control the behaviour? We should standardize.

feathr_project/test/test_config_loading.py
```diff
@@ -6,31 +6,26 @@
 
 def test_configuration_loading():
     """
-    Test CLI init() is working properly.
+    Test the configuration can be overwritten by envs
     """
     runner = CliRunner()
     with runner.isolated_filesystem():
 
         result = runner.invoke(init, [])
```
hangfei, line 15: we should moving away from using CLI to test. We can just using the test_user_workspace here.

src/main/scala/com/linkedin/feathr/offline/source/dataloader/hdfs/FileFormat.scala
```diff
@@ -41,7 +41,9 @@ object FileFormat {
       case p if p.endsWith(".orc") => ORC
       case p if p.endsWith(".avro.json") => AVRO_JSON
       case p if p.startsWith("jdbc:") => JDBC
-      case _ => PATHLIST}
+      case _ =>
+        if (ss.conf.get("spark.feathr.inputFormat","").nonEmpty) ss.conf.get("spark.feathr.inputFormat") else PATHLIST
+    }
```
hangfei, line 46: add comment for future dev

feathr_project/setup.py
```diff
@@ -35,6 +35,7 @@
         "pyyaml",
         "Jinja2",
         "tqdm",
+        "pyarrow",
```
hangfei, line 38: Let's put test deps into: https://stackoverflow.com/questions/15422527/best-practices-how-do-you-list-required-dependencies-in-your-setup-py.

hangfei (commented):
Overall LGTM.

jaymo001 (approved):

feathr_project/feathr/client.py
xiaoyongzhu: ya let me update. we should standardize

feathr_project/test/test_config_loading.py
xiaoyongzhu: agree

feathr_project/feathr/client.py
xiaoyongzhu: This is a good point and I should also document this in the code. To make it a bit clear:
- Job tags are for job metadata and it's not passed to the actual spark job (i.e. not visible to spark job), more like a platform related thing that Feathr want to add (currently job tags only have job output URL and job output format, ). They are carried over with the job and is visible to every Feathr client. Think this more like some customized metadata for the job which would be weird to be put in the spark job itself.
- Job arguments (or sometimes called job parameters)are the arguments which are command line arguments passed into the actual spark job. This is usually highly related with the spark job. In Feathr it's like the input to the scala spark CLI. They are usually not spark specific (for example if we want to specify the location of the feature files, or want to 
- Job configuration are like "configurations" for the spark job and are usually spark specific. For example, we want to control the no. of write parts for spark

Job configurations and job arguments (or sometimes called job parameters) have quite some overlaps (i.e. you can achieve the same goal by either using the job arguments/parameters vs. job configurations). But the job tags should just be used for metadata purpose.

hangfei (approved):

jaymo001 (approved):

PR #93, opened 2022-03-27
Elegant way to read output format

xiaoyongzhu:
Goal of this PR:
1. list all the supported cloud resources in README
2. a small refactoring to use an elegant way to read the output format

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ user_item_similarity = DerivedFeature(name="user_item_similarity",
 | Feature Registry             | Azure Purview                                                                                 |
 | Compute Engine               | Azure Synapse Spark Pools, Databricks                                                         |
 | Machine Learning Platform    | Azure Machine Learning, Jupyter Notebook                                                      |
-| File Format                  | Parquet, ORC, Avro, Delta, Lake                                                               |
+| File Format                  | Parquet, ORC, Avro, Delta Lake                                                                |
 
 ## Roadmap
 
--- a/README.md
+++ b/README.md
@@ -141,6 +141,18 @@ user_item_similarity = DerivedFeature(name="user_item_similarity",
                                       transform="cosine_similarity(user_embedding, item_embedding)")
 ```
 
+## Supported Format
+
+| Feathr component             | Cloud Integrations                                                                            |
+| ---------------------------- | --------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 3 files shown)


src/main/scala/com/linkedin/feathr/offline/generation/SparkIOUtils.scala
```diff
@@ -19,8 +19,11 @@ object SparkIOUtils {
   }
 
   def writeDataFrame( outputDF: DataFrame, path: String, parameters: Map[String, String] = Map()): DataFrame = {
-    // if the output format is set by spark configurations, we will use that as the job output format; otherwise use avro as default for backward compatibility
-    outputDF.write.mode(SaveMode.Overwrite).format(parameters.getOrElse(OUTPUT_FORMAT, "avro")).save(path)
+
+    val output_format = outputDF.sqlContext.getConf("spark.feathr.outputFormat", "avro")
```
jaymo001, line 23: We should update the documentation as well?

jaymo001: I saw it in the other PR. NVM.

README.md
```diff
@@ -141,6 +141,18 @@ user_item_similarity = DerivedFeature(name="user_item_similarity",
...
+
+| Feathr component             | Cloud Integrations                                                                            |
+| ---------------------------- | --------------------------------------------------------------------------------------------- |
+| Offline store – Object Store | Azure Blob Storage, Azure ADLS Gen2, AWS S3                                                   |
+| Offline store – SQL          | Azure SQL DB, Azure Synapse Dedicated SQL Pools (formerly SQL DW), Azure SQL in VM, Snowflake |
+| Online store                 | Azure Cache for Redis                                                                         |
+| Feature Registry             | Azure Purview                                                                                 |
+| Compute Engine               | Azure Synapse Spark Pools, Databricks                                                         |
+| Machine Learning Platform    | Azure Machine Learning, Jupyter Notebook                                                      |
+| File Format                  | Parquet, ORC, Avro, Delta Lake                                                                |
```
hangfei, line 154: this is a bit hard to make sense. probably defer this to a diff rb?

xiaoyongzhu: Ya I think it's like recording all the supported output format in the README file as @jaymo001 suggested, and we don't have that table for that hence also adding that table in the README file.

hangfei (approved):

PR #100, opened 2022-03-29
Use SQL expression based config by default

jaymo001:


feathr_project/test/test_feature_anchor.py
```diff
@@ -140,12 +140,12 @@ def test_agg_anchor_to_config():
...
             aggregationFeatures: {
                 source: nycTaxiBatchSource
-                key: [DOLocationID]
+                key.sqlExpr: [DOLocationID]
                 features: {
                     f_location_avg_fare: {
                         def: "cast_float(fare_amount)"
                         window: 90d
-                        agg: AVG
+                        aggregation: AVG
```
d4ve, line 148: What does this change do?

feathr_project/feathr/anchor.py
```diff
@@ -34,7 +34,7 @@ def to_feature_config(self) -> str:
         tm = Template("""
             {{anchor_name}}: {
                 source: {{source.name}}
-                key: [{{key_list}}]
+                key.sqlExpr: [{{key_list}}]
```
d4ve, line 37: In general this makes sense, but is there any effect on compatibility? Is the change motivated by performance?

jaymo001: No. It is compatible. Yes, It is motivated by performance.

feathr_project/test/test_feature_anchor.py
jaymo001: It should be aggregation instead of agg.

d4ve (approved):

xiaoyongzhu (approved):

PR #103, opened 2022-03-31
clean up unused functions in source utils

Yuqing-cat:


d4ve (commented):
This change might make the upcoming LI merge harder, and maybe should wait until after that. @jaymo001 WDYT?

hangfei:
> This change might make the upcoming LI merge harder, and maybe should wait until after that. @jaymo001 WDYT?

+1. Mentioned offline. We should defer this.

(you can hold on to this PR and re-check-in once Jay finishes his merge)

PR #109, opened 2022-04-03
Fix path parse issue

xiaoyongzhu:
This PR fixes an issue where if multiple files use multiple extensions (say one is using parquet and another one is using csv) the job will have partial failures. Previous tests covers this, but the spark job won't fail.

This PR also upgrades to the latest spark fashion of reading avro/orc files (otherwise the job will fail saying the setup is incorrect).

image


src/main/scala/com/linkedin/feathr/offline/source/dataloader/hdfs/FileFormat.scala
```diff
@@ -64,7 +65,20 @@ object FileFormat {
     // if we cannot detect the file type by extension, we will detect "spark.feathr.inputFormat" and use that as the option;
     // this is a global config (i.e. affecting all the inputs) so customers should use it as the last resort.
     // If this is not set, throw an exception (in `loadHdfsDataFrame()`)
-    if (existingHdfsPaths.head.endsWith(".parquet")) PARQUET else dataIOParameters.getOrElse(DATA_FORMAT, ss.conf.get("spark.feathr.inputFormat", AVRO)).toUpperCase
+    val p = existingHdfsPaths.head.toLowerCase()
+    p match {
```
jaymo001, line 69: Can we call getType instead of duplicating the code?

jaymo001 (approved):

xiaoyzhuLI (approved):

PR #117, opened 2022-04-10
Add doc for Feathr concepts for beginners

xiaoyongzhu:
This PR adds documentation to illustrate concepts for feathr beginners.


docs/concepts/feathr-concepts-for-beginners.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+## Illustration
+
+An illustration of the concepts that we are going to talk about is like this:
+![Feature Join Process](../images/observation_data.png)
+
+## Point in time joins and aggregations
+
+Assuming users are already familar with the "regular" joins, for example inner join or outer join, and in many of the use cases, we care about time, that is why in Feathr we provide a capability called Point in time Join (and with other time based aggregations). That is also a reason why we focus very much on time and need end users to specify the timestamp column.
```
jaymo001, line 113: Explicitly states that the feature timestamp is smaller than observation timestamp and this avoids information leak, i.e. using future data to train model at some time in the past, etc.

hangfei:
for observation_data.png: 

1. i don't feel a need to diff feature consumer or feature producer in the image as when they read the guide, they will be "the role". 
2. For feature consumer, you prob need quite different image here. If you want to combine feature consumer and producer together, you may want to add a example on how he/she consumers other's features here.
3. preprocessing should be on source and before features.
4. nit: It might be better actually give a example of source, like blog, example 2 as JDBC
5. the yellow feathr Project background doens't seem very informative

hangfei:
for feature_store_producer_consumer.png:
1. for first two steps, it should be define the features?
2. the arrow of the storage box should be labled with different people to demonstrate the idea of sharing
3. "put features in", "get features out"
4. feature store abstraction is a bit abstract. will "ELI5: feature store" or "how to understand feature store"

docs/concepts/feathr-concepts-for-beginners.md
```diff
@@ -0,0 +1,115 @@
...
+
+In this guide, we will cover the high level concepts for Feathr. Don't treat this as a user manual, instead treat this as blogpost to cover the highlevel motivations on why Feathr introduces those concepts.
+
+## What are `Observation` data, and why does Feathr need `key(s)`, `Anchor`, `Source`?
+
+In order to fully utilize Feathr's power, we need to understand the object models that Feathr is expecting.
+
+In Feathr, always think that there is some `Observation` dataset which is the central dataset that you will be using. The observation data set will usually have at least two columns: a timestamp column (indicating when this observation happens) and a column containing IDs, and with other possible fields
+
+For example, the `observation` dataset can be user click streams, credit card transactions, etc., while the data science team is trying to predict whether the user clicks on something, or whether the transcation is a fraud trasction, etc.
```
hangfei, line 17: we can just explain from the objective of DS(guess you are more familiar with that).

For example, as a DS, you want to build a model to do sth, then define obs ....

Then explain typically what was done without Feathr. But with Feathr, how things can be made easier.

docs/concepts/feathr-concepts-for-beginners.md
```diff
@@ -0,0 +1,115 @@
...
+
+In this case, how would we "link" the `observation` dataset, and the "additional dataset"? This is called [Feature Join](#why-does-feathr-need-feature-join-and-feature-query) in Feathr, but basically think this process as joining two tables.
+
+Since this is a table join process, we need to specify which `key(s)` that the join would happen. Those `keys` are usually some IDs, but can be others as well. In the above example, if we want to augment the user click stream data with user purchase history, we will use the user ID as `key`, so that the `click_stream` table and the `historical_buying` table can be joined together. Another case might be we want to get the historical blood pressure history for a patient, so `patient ID` will be used as key in that case. That's why you need to specify `keys` in Feathr `Feature`, becasue you will need to join your `Feature` with your `Observation Data` later on.
+
```
hangfei, line 24: let's use a example dataset. Users will get lost with so many words. Maybe a example is good here.

docs/concepts/feathr-concepts-for-beginners.md
```diff
@@ -0,0 +1,115 @@
...
+]
+
+request_anchor = FeatureAnchor(name="request_features",
+                               source=batch_source,
+                               features=features)
+```
+
+## Motivation on `Derived Feature`
+
+That sounds all good, but what if we want to share a feature, and others want to build additional features on top of that feature? Thats's why there is a concept in Feathr called `derived feature`, which allows you to calculate features based on other features, with certain transformation support. In practice, people can build features on top of other features and have a "feature chain".
```
hangfei, line 52: i feel we don't need to mention sharing here.

Maybe give two more examples of derived features. Like embedding multiplication. That will be easier to tell the story.

docs/concepts/feathr-concepts-for-beginners.md
```diff
@@ -0,0 +1,115 @@
...
+
+```python
+f_trip_time_distance = DerivedFeature(name="f_trip_time_distance",
+                                      feature_type=FLOAT,
+                                      input_features=[
+                                          f_trip_distance, f_trip_time_duration],
+                                      transform="f_trip_distance * f_trip_time_duration")
+```
+
+## Why does Feathr need `Feature Join` and `Feature Query`?
```
hangfei, line 64: we are retiring the concept of feature join for public APIs. instead, we are using get_offline features. So i believe you don't need to talk about join.

Instead, you should say: how to access your features

docs/concepts/feathr-concepts-for-beginners.md
```diff
@@ -0,0 +1,115 @@
...
+settings = ObservationSettings(
+    observation_path="wasbs://public@azurefeathrstorage.blob.core.windows.net/sample_data/green_tripdata_2020-04.csv",
+    event_timestamp_column="lpep_dropoff_datetime",
+    timestamp_format="yyyy-MM-dd HH:mm:ss")
+client.get_offline_features(observation_settings=settings,
+                            feature_query=feature_query,
+                            output_path=output_path)
+```
+
+## What is "materialization" in Feathr?
```
hangfei, line 88: How to access features from online storage

docs/concepts/feathr-concepts-for-beginners.md
```diff
@@ -0,0 +1,115 @@
...
+## Illustration
+
+An illustration of the concepts that we are going to talk about is like this:
+![Feature Join Process](../images/observation_data.png)
+
+## Point in time joins and aggregations
+
+Assuming users are already familar with the "regular" joins, for example inner join or outer join, and in many of the use cases, we care about time, that is why in Feathr we provide a capability called Point in time Join (and with other time based aggregations). That is also a reason why we focus very much on time and need end users to specify the timestamp column.
+
+For more details on how to utilize Feathr to perform point-in-time joins, refer to the [Point in Time Join Guide](../concepts/point-in-time-join.md)
```
hangfei, line 115: Add what's next for users. (quick start or full guide etc)

hangfei:
The guide is great. Thanks for adding the missing guide for new users.

jaymo001 (approved):

hangfei (approved):

jaymo001 (approved):

PR #119, opened 2022-04-11
Python client code to generate JDBC data source config

windoze:
Python client code to generate JDBC data source config


hangfei (commented):
Seems we are changing the config language here which are public APIs. 

For public APIs, we need a lot more attention. Let's sync offline with Jay anx Xiaoyong on this.

PR #120, opened 2022-04-12
[feahtr] Pull request guideline

hangfei:
What: create a pull request guideline for devs
Why: a established PR convention can expedite review process
How: PR guideline

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -119,6 +119,29 @@ batch_source = HdfsSource(
     timestamp_format="yyyy-MM-dd HH:mm:ss")                 # Supports various fromats inculding epoch
 ```
 
+## Defining Preprocessing on Source
+If you need some complex transformation that can be done with provided transformations, you can use source preprocessing
+to achieve your goal.
+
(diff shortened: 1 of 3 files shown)


jaymo001 (approved):

docs/README.md
```diff
@@ -119,6 +119,29 @@ batch_source = HdfsSource(
     timestamp_format="yyyy-MM-dd HH:mm:ss")                 # Supports various fromats inculding epoch
 ```
 
+## Defining Preprocessing on Source
+If you need some complex transformation that can be done with provided transformations, you can use source preprocessing
+to achieve your goal.
+
+The preprocessing takes in the DataFrame loaded by the source and then return a preprocessed DataFrame by the preprocessing
+function. The preprocessing function can't refer to other functions or dependencies that are not provided in the Spark cluster.
```
xiaoyongzhu, line 127: Can we specify the following?
1. it should be a single function; if calling other functions, that functions must be included in the preprocessing functions's body (i.e. the function is self contained)
2. the function should have a DF as an input, and need to return another df
3.

hangfei: discussed offline. will do 3 in later PR.

PR #101, opened 2022-03-30
Windoze/data source refactor

windoze:
Support other data source in DataSource.location

PR #124, opened 2022-04-12
Windoze/jdbc python

windoze:
This is the Python part corresponding to #101.
In #101, I added Scala code to handle JDBC sources using multiple sources need different auth credential, but the Python client still needs this update to let user do: 1. Create JdbcSource, 2. Pass required parameter to Spark job.
This PR adds a new Spark job argument `--system-properties`, which is used to pass secrets from Python client to Spark job, as we shall not store secrets directly inside the config files. The key of each entry is the data source name with `_USER`/`_PASSWORD` or `_TOKEN` suffices, depends on the data source auth type, and the value is taken from the current environment variables with the corresponding key from the Python client side.


hangfei:
Hi @windoze , we want to retrospectively follow https://github.com/linkedin/feathr/blob/main/docs/dev_guide/pull_request_guideline.md. Could you create a github issue for your PR? Also let's sync first to align on the technical direction with @xiaoyongzhu

windoze:
This PR addresses #102

feathr_project/feathr/_abc.py
```diff
@@ -30,6 +30,7 @@ def submit_feathr_job(self, job_name: str, main_jar_path: str,  main_class_name:
             arguments (str): all the arugments you want to pass into the spark job
             job_tags (str): tags of the job, for exmaple you might want to put your user ID, or a tag with a certain information
             configuration (Dict[str, str]): Additional configs for the spark job
+            properties (Dict[str, str]): Additional System Properties for the spark job
```
xiaoyongzhu, line 33: What's this system property for?

xiaoyongzhu:
I'm not sure if I understand the background of this PR, but I feel this 

@windoze maybe you can add a bit more description?

PR #138, opened 2022-04-19
Fix corner case when registering derived features

xiaoyongzhu:
Currently if a derived feature is only based on anchor features, it will not be registered correctly. This PR solves the issue (also test updated to prevent this issue in the future)

### DIFF
--- a/feathr_project/test/test_fixture.py
+++ b/feathr_project/test/test_fixture.py
@@ -127,7 +127,7 @@ def snowflake_test_setup(config_path: str):
 def registry_test_setup(config_path: str):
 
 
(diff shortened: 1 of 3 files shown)


feathr_project/test/test_fixture.py
```diff
@@ -127,6 +127,10 @@ def snowflake_test_setup(config_path: str):
 def registry_test_setup(config_path: str):
 
 
+    # use a new project name every time to make sure if all features are registered correctly no
```
jaymo001, line 130: nit: typo? 'correctly no'

hangfei (approved):

jaymo001 (approved):

PR #139, opened 2022-04-19
Windoze/udf improve

windoze:
Use `cloudpickle` to save and load UDF, to support complex callable objects such as functions with dependencies, lambdas, callable classes, etc.
Choosing `cloudpickle` is because this package is shipped within PySpark so we don't need to involve additional dependencies.

### DIFF
--- a/feathr_project/feathr/_preprocessing_pyudf_manager.py
+++ b/feathr_project/feathr/_preprocessing_pyudf_manager.py
@@ -5,6 +5,7 @@ from typing import List, Optional, Union
 import pickle
 from jinja2 import Template
 from feathr.source import HdfsSource
+from pyspark import cloudpickle
 
 # Some metadata that are only needed by Feathr
 FEATHR_PYSPARK_METADATA = 'generated_feathr_pyspark_metadata'
@@ -16,7 +17,8 @@ FEATHR_PYSPARK_DRIVER_FILE_NAME = 'feathr_pyspark_driver.py'
 FEATHR_PYSPARK_DRIVER_TEMPLATE_FILE_NAME = 'feathr_pyspark_driver_template.py'
 # Feathr provided imports for pyspark UDFs all go here
 PROVIDED_IMPORTS = ['\nfrom pyspark.sql import SparkSession, DataFrame\n'] + \
-                   ['from pyspark.sql.functions import *\n']
+                   ['from pyspark.sql.functions import *\n'] + \
+                   ['from pyspark import cloudpickle\n__feathr_user_functions={}\n']
 
 
 class _PreprocessingPyudfManager(object):
@@ -27,8 +29,10 @@ class _PreprocessingPyudfManager(object):
         """When the client build features, UDFs and features that need preprocessing will be stored as metadata. Those
         metadata will later be used when uploading the Pyspark jobs.
         """
-        # feature names concatenated to UDF map
-        # for example, {'f1,f2,f3': my_udf1, 'f4,f5':my_udf2}
+        # feature names concatenated to UDF callable object map
+        # the UDF callable is stored in `__feathr_user_functions` dictionary with source name as the key
+        # so the map is like:
+        # 'f1,f2': __feathr_user_functions['source1']
         feature_names_to_func_mapping = {}
         # features that have preprocessing defined. This is used to figure out if we need to kick off Pyspark
         # preprocessing for requested features.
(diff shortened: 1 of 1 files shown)


hangfei:
Please follow: https://github.com/linkedin/feathr/blob/main/docs/dev_guide/pull_request_guideline.md

windoze:
Updated PR description.

hangfei:
If a function depends on a chain of function, or a function depdends on a library that is huge, how does the pickle work? Will it still work(recursively)? Will it the function pickle become too huge?

feathr_project/feathr/_preprocessing_pyudf_manager.py
```diff
@@ -39,12 +43,12 @@ def build_anchor_preprocessing_metadata(anchor_list, local_workspace_dir):
                 continue
             preprocessing_func = anchor.source.preprocessing
             if preprocessing_func:
-                _PreprocessingPyudfManager.persist_pyspark_udf_to_file(preprocessing_func, local_workspace_dir)
+                _PreprocessingPyudfManager.persist_pyspark_udf_to_file(anchor.source.name, preprocessing_func, local_workspace_dir)
```
hangfei, line 46: i use a separate UDF file for ease of management and readability since i am writing the source code.

Since you are using pickle now, you can just put it all in the feature_names_to_func_mapping in one shot.

hangfei: after you combine them, you probably don't need source name any  more. source name is not stable and is less preferred.

windoze:
@hangfei yes it will, cloudpickle serialize object recursively by value, the process stops at the module boundary, and will trigger module importing on deserialization when needed, so the only limitation appears when some local modules are not available on the remote side(there is a [workaround](https://pypi.org/project/cloudpickle/) but needs few manual steps).
The same process is used by PySpark, it uses cloudpickle to transfer compute tasks between drivers and executors.

The package size will not be significant usually, but some caution needs to be taken, i.e. if you create a 1M elements list in your UDF, the list will also be packed into the output byte array and increase its size dramatically, but this issue is actually inevitable, not matter what solution do you use.

hangfei:
> @hangfei yes it will, cloudpickle serialize object recursively by value, the process stops at the module boundary, and will trigger module importing on deserialization when needed, so the only limitation appears when some local modules are not available on the remote side(there is a [workaround](https://pypi.org/project/cloudpickle/) but needs few manual steps). The same process is used by PySpark, it uses cloudpickle to transfer compute tasks between drivers and executors.
> 
> The package size will not be significant usually, but some caution needs to be taken, i.e. if you create a 1M elements list in your UDF, the list will also be packed into the output byte array and increase its size dramatically, but this issue is actually inevitable, not matter what solution do you use.

module is equal to a lib?

Could you provide a user guide for this part? better with examples on their own dependencies, code that they import from their teammates or some third party libs.

windoze:
I think we can just add a note that refers to the official cloudpickle document, no need to add too many our own words or comments.

feathr_project/feathr/_preprocessing_pyudf_manager.py
windoze: Here is the thing, if we put pickled code directly inside `feature_names_to_func_mapping`, there will be duplicates if one UDF is used by multiple places, and even worse, if there is any one-shot logics inside of this UDF, it will be executed multiple times, that is semantically different than the original implementation and may even cause bugs.

hangfei:
> d a few lines in the de

Do you mean in addition to user guide, we can mention pickle in dev guide? I feel you are right. We should do both here but with different emphasis. User guide should not mention pickle. Dev guide should mention pickle.

feathr_project/feathr/_preprocessing_pyudf_manager.py
windoze: You're right, this part is not necessary, removed.

xiaoyongzhu (approved):

xiaoyongzhu:
Not sure why the envs are not read correctly (probably because this PR is opened earlier than the PR which enables forked repo test)

hangfei:
Release candidate for v0.4.0

hangfei:
Please fix the tests.

hangfei (approved):

xiaoyongzhu:
The test fails because of credentials not found. I'm merging it to see the final CI, and will keep guys updated.

PR #141, opened 2022-04-19
[Ready to Merge] Enable key vault as secret source

Yuqing-cat:
- add key vault dependency
- create a credential client class to init key vault secret client
- add redis sample in credential util which can replace current _construct_redis_client()
- will add more functions and tests if above change is acceptable.


hangfei:
Please follow https://github.com/linkedin/feathr/blob/main/docs/dev_guide/pull_request_guideline.md.

Yuqing-cat:
What changes to make and why you are making these changes:
- Add akv_client to support reading secrets from Azure Key Vault.
- Key vault logic will be triggered only when environment variable is not set and KEY_VAULT__NAME is given. (envutil)
- Update ARM Template to store REDIS_PASSWORD rather than CONN for better akv usage 

2. How are you going to achieve your goal Describe what testings you have done, for example, performance testing etc.
- So far it will perform exactly the same as it was unless and secret variables are removed and "KEY_VAULT__NAME" is set.

xiaoyongzhu:
@Yuqing-cat let's talk a bit about the scope of this work (probably need a bit bigger scope to cover the spark runtime as well)

feathr_project/feathr/_envvariableutil.py
```diff
@@ -59,5 +61,8 @@ def get_environment_variable(variable_key):
         password = os.environ.get(variable_key)
         if not password:
             logger.info(variable_key +
-                        ' is not set in the environment variables.')
+                        ' is not set in the environment variables. Try Key Vault value.')
```
jainr, line 64: When an environment variable is not found, is the expectation for them to set the environment variable with keyVaultName so it is already set?
If it's latter, I suggest modifying the log message to ".....' is not set in the environment variables, fetching the value from Key Vault')

feathr_project/feathr/_envvariableutil.py
```diff
@@ -59,5 +61,8 @@ def get_environment_variable(variable_key):
...
+            akv_name = os.env.get("KEY_VAULT__NAME")
```
jainr, line 65: [nitpick] Suggest changing it to KEY_VAULT_NAME , same number of underscores.

feathr_project/feathr/akv_client.py
```diff
@@ -0,0 +1,22 @@
+from azure.keyvault.secrets import SecretClient
+from azure.identity import DefaultAzureCredential
+from loguru import logger
+from azure.core.exceptions import ResourceNotFoundError
+
+class akvClient:
```
jainr, line 6: [nitpick] - Suggest changing it to more descriptive name AzureKeyVaultClient, the way we have for PurviewClient etc.

jainr:
Is KV only storing Redis  Credetials for Online store?  Can we store other credentials too in KV that will unify the implementation when other environment variables are not set.

feathr_project/feathr/_envvariableutil.py
Yuqing-cat: was intended to fit the naming pattern of yaml config attributes. But agree, we can use KEY_VAULT_NAME for current functions

feathr_project/feathr/_envvariableutil.py
Yuqing-cat: updated as recommended.

xiaoyongzhu (approved):

windoze (approved):

PR #151, opened 2022-04-21
Adding Databricks quickstart guide

xiaoyongzhu:
This PR aims at providing a "one-click" run experience for databricks clusters, ideally without any configuration at all. To achieve this:
1. Added a sample notebook for databricks cluster. 
2. In order to achieve "zero configuration", we've deleted many contents in the notebook (like registry connections etc.)

Also this comes with a usability fix, when the job fails, the stacktrace of the job will be print out so that end users don't have to go to spark jobs to look at the failure and error message.

A separate PR will be sent to highlight this notebook in documentation.


hangfei:
Release candidate for v0.4.0

hangfei:
Release candidate for v0.4.0

hangfei (dismissed):

hangfei (approved):

jaymo001 (approved):

PR #159, opened 2022-04-25
Fix Redis suffix problem for snowflake

Yuqing-cat:
PR includes:
- create a common `get_online_test_table_name()` in `test_fixture`
- replace below with `online_test_table = get_online_test_table_name('nycTaxiCITable')`
```
now = datetime.now()
online_test_table = ''.join(['nycTaxiCITable', '_', str(now.minute), '_', str(now.second)])
```
- add table suffix logic in `snowflake_e2e`

### DIFF
--- a/feathr_project/test/test_fixture.py
+++ b/feathr_project/test/test_fixture.py
@@ -256,4 +256,4 @@ def registry_test_setup(config_path: str):
 def get_online_test_table_name(table_name: str):
     # use different time for testing to avoid write conflicts
     now = datetime.now()
-    return ''.join([table_name,'_', str(now.minute), '_', str(now.second)])
+    return '_'.join([table_name, str(now.minute), str(now.second)])


xiaoyongzhu (dismissed):

xiaoyongzhu:
Calling get_online_test_table_name() in the same test from different function will yield different results, which needs to be addressed

hangfei:
Release candidate for v0.4.0

feathr_project/test/test_fixture.py
```diff
@@ -251,3 +252,8 @@ def add_new_dropoff_and_fare_amount_column(df: DataFrame):
     random.shuffle(derived_feature_list)
     client.build_features(anchor_list=[agg_anchor, request_anchor], derived_feature_list=derived_feature_list)
     return client
+
+def get_online_test_table_name(table_name: str):
+    # use different time for testing to avoid write conflicts
+    now = datetime.now()
+    return ''.join([table_name,'_', str(now.minute), '_', str(now.second)])
```
windoze, line 259: '_'.join([table_name, str(now.minute), str(now.second)])

Yuqing-cat: Agree. updated.

xiaoyongzhu (approved):

windoze (approved):

PR #170, opened 2022-04-26
Setting up initial maven publish dependencies/configs

blee1234:
Adding[ sbt-sonatype support](https://github.com/xerial/sbt-sonatype), which is a third-party plugin that essentially supports a one-line command to publish to maven.

Had to alter some of the comments, as they were causing scala-doc issues that prevented bundling. In addition, organization name seemed to have been incorrectly set before, as it doesn't follow syntax presented [here](https://www.scala-sbt.org/1.x/docs/Using-Sonatype.html), and was causing release issues.


xiaoyongzhu (dismissed):

xiaoyongzhu:
Test failure should be unrelated with this PR. the issue was because an earlier PR which @windoze is fixing

xiaoyongzhu:
@blee1234 there are a few issues with the recent CI pipeline (https://github.com/linkedin/feathr/issues/174) which we just had mitigation in the last a few commits. Can you merge main back to your branch to run all the tests?

xiaoyongzhu (approved):

xiaoyongzhu:
I'm not sure if we still want to keep this jar as 0.1.x or should we bump this to 0.4.x which is more aligned with the python package. I'll leave this discussion to @jaymo001

jaymo001 (approved):

Issue #167, opened 2022-04-26
UDF E2E tests failed


windoze:
After merging #139, UDF E2E tests failed as the PR changed to use `pyspark.cloudpickle` instead of `inspect.getsource()`.

The root cause is that the `cloudpickle` by default will **not** pickle modules by value, and the PyTest doesn't run test code in `__main__` as normal python execution flow, that means all UDFs defined in the test code belongs to a different module than `__main__`, this leads to a mysterious `ModuleNotFound` exception when pickled UDFs were used in the remote PySpark job, as it cannot find the module where the pickled function belongs to, because the test module doesn't exist on the remote side.

Since version 2.0, `cloudpickle` resolves this issue by involving a new `register_pickle_by_value(module)` function, but unfortunately this function doesn't exist in the `cloudpickle` bundled with PySpark, which stays at 1.6 for years.

This issue actually will **not** happen in typical usage, e.g. in a Python REPL env or Jupyter notebook, as they all run current code in `__main__` module, it only affects PyTest because its running model.

One possible solution is to use the new `cloudpickle` but it introduces new dependencies which must be installed on the remote Spark cluster manually.

The other solution is to let user package their UDF into a separated Python file and submit this file along with the Spark job, but this is not compatible with current test code thus needs a major restructuring.

The workaround is pretty straightforward, changing the module name to `__main__` at the very beginning of the E2E test script, it looks like a hack but actually does the job pretty well.

PR #177, opened 2022-04-27
Fix e2e tests

windoze:
#167  fix

### DIFF
--- a/feathr_project/feathr/_preprocessing_pyudf_manager.py
+++ b/feathr_project/feathr/_preprocessing_pyudf_manager.py
@@ -1,6 +1,7 @@
 import inspect
 import os
 from pathlib import Path
+import sys
 from typing import List, Optional, Union
 import pickle
 from jinja2 import Template
@@ -37,15 +38,23 @@ class _PreprocessingPyudfManager(object):
         feature_names_to_func_mapping = {}
         # features that have preprocessing defined. This is used to figure out if we need to kick off Pyspark
         # preprocessing for requested features.
-        features_with_preprocessing = []
+        features_with_preprocessing = {}
+        dep_modules = set()
         for anchor in anchor_list:
             # only support batch source preprocessing for now.
             if not isinstance(anchor.source, HdfsSource):
                 continue
             preprocessing_func = anchor.source.preprocessing
             if preprocessing_func:
+                # Record module needed by all preprocessing_func
+                for feature in anchor.features:
+                    # Record module file name in metadata
+                    features_with_preprocessing[feature.name] = sys.modules[preprocessing_func.__module__].__file__
+                    # Record module name except __main__
+                    if preprocessing_func.__module__ != "__main__":
+                        dep_modules.add(preprocessing_func.__module__)
                 feature_names = [feature.name for feature in anchor.features]
-                features_with_preprocessing = features_with_preprocessing + feature_names
+                # features_with_preprocessing = features_with_preprocessing + feature_names
                 feature_names.sort()
                 string_feature_list = ','.join(feature_names)
                 feature_names_to_func_mapping[string_feature_list] = "cloudpickle.loads(%s)" % cloudpickle.dumps(preprocessing_func)
@@ -53,7 +62,7 @@ class _PreprocessingPyudfManager(object):
         if not features_with_preprocessing:
             return
 
-        _PreprocessingPyudfManager.write_feature_names_to_udf_name_file(feature_names_to_func_mapping, local_workspace_dir)
+        _PreprocessingPyudfManager.write_feature_names_to_udf_name_file(feature_names_to_func_mapping, dep_modules, local_workspace_dir)
 
         # Save necessary preprocessing-related metadata locally in your workspace
         # Typically it's used as a metadata for join/gen job to figure out if there is preprocessing UDF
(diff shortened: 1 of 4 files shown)


hangfei:
Please follow https://github.com/linkedin/feathr/blob/main/docs/dev_guide/pull_request_guideline.md so users can track and refer to this in the future.

windoze:
PR description updated.

xiaoyongzhu (approved):

xiaoyongzhu (dismissed):

hangfei:
Could you explain this part: `Usually cloudpickle works well if the UDF is defined in __main__`

hangfei:
If the user has a UDF that refer to another function in a sub module, does this work? 

I am reading about https://stackoverflow.com/questions/2121874/python-pickling-after-changing-a-modules-directory. Seem it might not work as well.

xiaoyongzhu:
> https://stackoverflow.com/questions/2121874/python-pickling-after-changing-a-modules-directory

Regular Pickle won't work for sure, cloudpickle should work from what I chatted with @windoze

windoze:
> 

This PR uses the same method as PySpark does for UDFs, so it imposes the same restriction:

Functions must don't refer to anything in another **private** module, i.e. a module doesn't exist on the remote Spark side.

And this PR added an improvement:

The module that contains the UDF function, will be transferred to the remote side, so the UDF functions can be located in separated module. The purpose of this improvement is to support PyTest, but it can be useful to users.

feathr_project/feathr/feathr_pyspark_driver_template.py
```diff
@@ -82,3 +84,24 @@ def submit_spark_job(feature_names_funcs):
     py4j_feature_job.mainWithPreprocessedDataFrame(job_param_java_array, new_preprocessed_df_map)
     return None
 
+def decode_file(filename, encoded):
+    import base64
+    import os.path
```
hangfei, line 89: move import to top.
add documentation

feathr_project/feathr/feathr_pyspark_driver_template.py
```diff
@@ -82,3 +84,24 @@ def submit_spark_job(feature_names_funcs):
...
+            start_dir = os.path.dirname(this_filename)
+        except:
+            start_dir = os.path.curdir
+        if start_dir not in sys.path:
+            sys.path.append(start_dir)
+        output_name = os.path.join(start_dir, filename)
+        print("Write file %s ..." % output_name)
+        with open(output_name, "w+b") as f:
+            f.write(content)
+        print("File %s written." % output_name)
```
hangfei, line 105: this will show up in user facing prints right? Let's not add them here.

feathr_project/feathr/_preprocessing_pyudf_manager.py
```diff
@@ -154,3 +184,16 @@ def prepare_pyspark_udf_files(feature_names: List[str], local_workspace_dir):
 
             py_udf_files = [pyspark_driver_path]
         return py_udf_files
+
+import base64
```
hangfei, line 188: move import to top

feathr_project/feathr/_preprocessing_pyudf_manager.py
```diff
@@ -59,47 +78,57 @@ def build_anchor_preprocessing_metadata(anchor_list, local_workspace_dir):
...
-                print("".join(lines), file=handle)
-
-    @staticmethod
-    def write_feature_names_to_udf_name_file(feature_names_to_func_mapping, local_workspace_dir):
+    def write_feature_names_to_udf_name_file(feature_names_to_func_mapping, dep_modules, local_workspace_dir):
         """Persist feature names(sorted) of an anchor to the corresponding preprocessing function name to source path
         under the local_workspace_dir.
         """
+        file_contents = []
+        for m in dep_modules:
```
hangfei, line 86: avoid acronym. m -> dep_module

feathr_project/feathr/_preprocessing_pyudf_manager.py
```diff
@@ -59,47 +78,57 @@ def build_anchor_preprocessing_metadata(anchor_list, local_workspace_dir):
...
+            """
+            The `feature_names_funcs` map contents a `cloudpickle.loads` statement that deserialize the
+            UDF function, due to the implementation limitation, cloudpickle shipped with PySpark cannot
+            actually serialize the function body if the function is in a module other than `__main__`,
+            which causes PyTest failed to run the UDF E2E test as it starts the test code in a local module
+            instead of `__main__`.
+            The solution is to bundle addition module with the main file.
+            First we encode module file in base64 and paste it in following format
```
hangfei, line 94: document why we need base64 encoding for future dev.

feathr_project/feathr/_preprocessing_pyudf_manager.py
```diff
@@ -27,29 +31,44 @@ def build_anchor_preprocessing_metadata(anchor_list, local_workspace_dir):
...
+                        # In case the module name cannot be retrieved
+                        features_with_preprocessing[feature.name] = sys.modules[preprocessing_func.__module__].__file__
+                    except:
+                        pass
+                    # Record module name except __main__
+                    if preprocessing_func.__module__ != "__main__":
+                        dep_modules.add(preprocessing_func.__module__)
                 feature_names = [feature.name for feature in anchor.features]
-                features_with_preprocessing = features_with_preprocessing + feature_names
+                # features_with_preprocessing = features_with_preprocessing + feature_names
```
hangfei, line 63: remove this?

hangfei (commented):
Could you add a dev guide for this? It might be hard for future dev to pick this up just by reading the code.

xiaoyongzhu (approved):

xiaoyongzhu:
Talked with @windoze offline, let’s NOT merge his PR in as there might be compatible Python issues, which will be very impactful for end user experience

more specifically, his implementation can solve many other issues (like importing UDF from other modules, support chaining of UDFs etc.). however, one big disadvantage is, it requires the python version of the local client version the same with the python version on the cluster. Some times it’s not the case and sometimes it’s not, like client 3.9 + cluster 3.8 can work, but client 3.7+cluster 3.8 won’t work, and we don’t know in what kind of situations it will work or not

xiaoyongzhu:
I believe this is the root cause:
https://github.com/cloudpipe/cloudpickle/issues/404#issuecomment-1094965687

```
When reducing a code object for pickling, cloudpickle stores it as a list of arguments to CodeType constructor: [code](https://github.com/cloudpipe/cloudpickle/blob/f758eb34d1b3285dc582b73bfd8df4c47ed4fc68/cloudpickle/cloudpickle_fast.py#L245). As you can see in the linked code, the list of those arguments changes between Python versions. Therefore pickling using one version of Python and unpickling using a different version does not work.

It could be resolved by creating a custom function to be used in reconstructing code objects, and having it deal with different Python source versions. However, I'm not sure if this would be a helpful improvement, or a false promise if passing pickled objects between different Python versions (or different cloudpickle versions, for that matter) is not expected to work anyway. Looking forward for the thoughts of cloudpickle maintainers on this issue.

```

xiaoyongzhu:
Also some other pointers: https://github.com/facebookincubator/submitit/issues/39

xiaoyongzhu:
One way of doing this (given cloudpickle's limitation) might be:
1. cloudpickle user's UDF
2. create a new module in the background, unpickle the UDF locally
3. pickle the new module with cloudpickle, this will be version compatible per the issue mentioned (https://github.com/facebookincubator/submitit/issues/39)

Yuqing-cat:
Hi @windoze , is this PR still active?

windoze:
> Hi @windoze , is this PR still active?

No it's not, the whole design is still under reconsideration, we can close it for now.

PR #180, opened 2022-04-27
Use Databricks CLI instead of REST APIs

xiaoyongzhu:
Currently we are using REST APIs to submit databricks jobs. This is definitely less than ideal and hard to maintain (as there are a lot of hard coded paths etc.). Also it cause some troubles when downloading the files during testing.

Although there's no databricks python SDK, we can use databricks-cli as a python SDK since it's written in Python. 
https://github.com/databricks/databricks-cli

For example, copying file can be achieved using this API:
https://github.com/databricks/databricks-cli/blob/master/databricks_cli/dbfs/cli.py#L118

This issue is opened to change the implementation to use databricks-cli as a python SDK to submit jobs, get files, etc. to make the code more elegant. This also should solve the periodic failure that databricks test might have (when downloading files)

This PR introduce `databricks-cli` package which is Apache 2 (https://github.com/databricks/databricks-cli/blob/master/LICENSE.txt).

credit goes to @windoze for this idea.

This PR also introduce a small improvement: when databricks job fails, it will print out job link so that users can view the error message. Ideally all the job log should be retrieved and @windoze is working on that improvement.

### DIFF
--- a/feathr_project/feathr/_databricks_submission.py
+++ b/feathr_project/feathr/_databricks_submission.py
@@ -154,11 +154,12 @@ class _FeathrDatabricksJobLauncher(SparkJobLauncher):
             # see if we can parse the returned result
             self.res_job_id = result['run_id']
         except:
-            logger.error("submitting to Databricks cluster failed. Message returned from Databricks: {}", result)
+            logger.error("Submitting Feathr job to Databricks cluster failed. Message returned from Databricks: {}", result)
             exit(1)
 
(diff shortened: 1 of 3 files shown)


hangfei (approved):

hangfei:
LGTM. Just curious to know what's the diff between CLI and python SDK in the context of Databrics? It seems quite diff in our CLI and SDK but it's different in databrics CLI.

Yuqing-cat (approved):

hangfei (approved):

PR #190, opened 2022-04-29
[feathr] Move pyspark into test requires

hangfei:
Most users with Azure already have pyspark dependency and pyspark is pretty large. So take it out from dependency and move to test for now.


xiaoyongzhu (approved):

hangfei:
This is dep move change and blocks the client. So commit with one approval.

PR #197, opened 2022-04-29
Detect returned error and print accordingly

xiaoyongzhu:
Previously, when we detect a databricks job is failed, we will print out the logs assuming there are `error` and `error_trace` field. This is probably not true in some of the cases so we need to make those print out in a conditional way.

### DIFF
--- a/feathr_project/feathr/_databricks_submission.py
+++ b/feathr_project/feathr/_databricks_submission.py
@@ -182,8 +182,8 @@ class _FeathrDatabricksJobLauncher(SparkJobLauncher):
             elif status in {'INTERNAL_ERROR', 'FAILED', 'TIMEDOUT', 'CANCELED'}:
                 result = RunsApi(self.api_client).get_run_output(self.res_job_id)
                 # See here for the returned fields: https://docs.microsoft.com/en-us/azure/databricks/dev-tools/api/2.0/jobs#--response-structure-8
-                # print out logs and stack trace if the job is failed
-                logger.error("Feathr job is failed. Please visit this page to view error message: {}", self.job_url)
+                # print out logs and stack trace if the job has failed
+                logger.error("Feathr job has failed. Please visit this page to view error message: {}", self.job_url)
                 if "error" in result:
                     logger.error("Error Code: {}", result["error"])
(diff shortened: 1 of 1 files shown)


feathr_project/feathr/_databricks_submission.py
```diff
@@ -184,8 +184,10 @@ def wait_for_completion(self, timeout_seconds: Optional[int] = 600) -> bool:
                 # See here for the returned fields: https://docs.microsoft.com/en-us/azure/databricks/dev-tools/api/2.0/jobs#--response-structure-8
                 # print out logs and stack trace if the job is failed
                 logger.error("Feathr job is failed. Please visit this page to view error message: {}", self.job_url)
```
hangfei, line 186: has failed

feathr_project/feathr/_databricks_submission.py
```diff
@@ -184,8 +184,10 @@ def wait_for_completion(self, timeout_seconds: Optional[int] = 600) -> bool:
...
-                logger.error("Error Code: {}", result["error"])
-                logger.error("{}", result["error_trace"])
+                if "error" in result:
```
hangfei, line 187: if result["error"]?

feathr_project/feathr/_databricks_submission.py
```diff
@@ -184,8 +184,10 @@ def wait_for_completion(self, timeout_seconds: Optional[int] = 600) -> bool:
...
+                    logger.error("Error Code: {}", result["error"])
+                if "error_trace" in result:
```
hangfei, line 189: if result["error_trace"]

feathr_project/feathr/_databricks_submission.py
xiaoyongzhu: that won't work (it will throw out exception). `if "key" in dict` is recommended way to detect key in python dict

feathr_project/feathr/_databricks_submission.py
xiaoyongzhu: same above

jaymo001 (approved):

PR #205, opened 2022-05-02
Setting up Maven

blee1234:
Adding[ sbt-sonatype support](https://github.com/xerial/sbt-sonatype), which is a third-party plugin that essentially supports a one-line command to publish to maven.

Had to alter some of the comments, as they were causing scala-doc issues that prevented bundling. 

In addition, organization name seemed to have been incorrectly set before, as it doesn't follow syntax presented [here](https://www.scala-sbt.org/1.x/docs/Using-Sonatype.html), and was causing release issues.

Bumped dependencies, as there were some blocking vulnerabilities.

Bumped feathr scala version to match feathr python version.

Also added a plugin to help identify dependencies.

Original Request. 
https://github.com/linkedin/feathr/pull/170

Re-review with a cleaner up commit history.


xiaoyongzhu:
@blee1234 can you merge main back to your branch? the test failure should be fixed by @hangfei

xiaoyongzhu (approved):

jaymo001 (approved):

PR #232, opened 2022-05-06
Define registered feature user story and test cases

jaymo001:


Yuqing-cat:
Hi @jaymo001 , is this PR still active?

Yuqing-cat:
@blrchen , please take a look and see if it's useful for your test case refactor.

PR #233, opened 2022-05-06
Changing illegal key vault secret name

adityakaul:
This PR fixes a typo in the ARM template shown in the Feathr quickstart. Currently, the deploy to azure button fails to deploy because the secret name REDIS_PASSWORD is illegal .Secrets can only contain alphanumeric characters and dashes. Changing it to REDIS-PASSWORD to match the rest of the template. This secret is not used in the jupyter notebook shown in quick start, so does not need to be changed there.

### DIFF
--- a/docs/how-to-guides/azure_resource_provision.json
+++ b/docs/how-to-guides/azure_resource_provision.json
@@ -116,7 +116,7 @@
                 },
                 {
                     "type": "Microsoft.KeyVault/vaults/secrets",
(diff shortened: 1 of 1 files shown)


Yuqing-cat (approved):

hangfei:
Could you help review change? @xiaoyongzhu

xiaoyongzhu (approved):

xiaoyongzhu:
Thanks for the fix @adityakaul !

Issue #236, opened 2022-05-09
Missing '[]' for key list


windoze:
The `key` field should be an array instead of a comma separated string.
https://github.com/linkedin/feathr/blob/ff66b2a3cd4966f8e79aa0e7f92b705f1eede46d/feathr_project/feathr/query_feature_list.py#L24

Corresponding Scala code reading this value is at:
https://github.com/linkedin/feathr/blob/ff66b2a3cd4966f8e79aa0e7f92b705f1eede46d/src/main/scala/com/linkedin/feathr/offline/config/FeatureJoinConfig.scala#L56

PR #237, opened 2022-05-09
#236 Missing brackets

windoze:
https://github.com/linkedin/feathr/issues/236

Missing brackets in the feature join config template.


xiaoyongzhu (approved):

hangfei (approved):
could we also add a test to cover this?

Issue #174, opened 2022-04-26
CI test does not work as expected


xiaoyongzhu:
There are two issues:
1. After @jainr 's PR, the CI code in main branch will not run
2. For forked repository, the test will only run once and won't run after further changes (reported by @windoze )
3. more comments below by other folks

Yuqing-cat:
My case is you need to unlabel and relabel the PR as "**safe to test**" to trigger the right action. And it will trigger a normal GitHub action (no secret) after further changes, which causes some fail records in PR.

jainr:
https://github.com/linkedin/feathr/pull/239

PR #246, opened 2022-05-11
Fixing CI Pipeline

jainr:
In the current workflow, the tests were not executing for subsequent commits and the latest code was not getting checked out. This PR fixes those two issues as mentioned here https://github.com/linkedin/feathr/issues/174


xiaoyongzhu:
@jainr can you also make sure we combine the two test yaml file? Seems this change won't run the code on main

PR #257, opened 2022-05-15
Improve error message when path is not supported

xiaoyongzhu:
There are cases where customers might put wasb/s3/etc. paths in databricks and wish Feathr can download those file and upload them to databricks. This is not supported by design, so adding warning & error message for that.


hangfei:
add pr description

hangfei (approved):

feathr_project/feathr/_databricks_submission.py
```diff
@@ -85,31 +85,35 @@ def upload_or_get_cloud_path(self, local_path_or_http_path: str):
         elif src_parse_result.scheme.startswith('dbfs'):
             # passed a cloud path
             logger.debug(
```
blrchen, line 87: Should this be a warning? 

For error message, it's better to tell what is supported too.

blrchen (approved):

hangfei (approved):

Issue #227, opened 2022-05-06
Add validation for feature name


hangfei:
Feature name should follow some naming conventions and rules otherwise it will break certain underlying engine, like spark or storage.

lhayhurst:
As per the comment on line 18 of setup.py:

>Only alphabet, numbers, and '_' are allowed in the name. It can not start with numbers. Note that '.' is NOT ALLOWED!

Are these all the validation rules needed?

lhayhurst:
I banged out a PR on my fork; please let me know if this is the sort of validation you were thinking of? Of so, I can issue a PR against the main repo.

 https://github.com/lhayhurst/feathr/pull/1

lhayhurst:
@hangfei  @xiaoyongzhu  Hello, just bumping this issue for visibility. Cheers

hangfei:
Hi @lhayhurst , the notifications are buried in my mailbox so i didn't see it earlier. Sry about the late reply.

Thanks for your interest on the project!

For the issue, yes. that's the only rules right now.

The PR looks good to me and only some minor comments.

@chinmaytredence  who are interested on leveraging the feature name validation work.

lhayhurst:
@hangfei   @chinmaytredence  PR is up https://github.com/linkedin/feathr/pull/258/files

chinmaytredence:
@lhayhurst thanks for this contribution.
@hangfei this would help us for create entity api. I think the exception message could also be sent as response while exposing through fast api

hangfei:
@chinmaytredence Yeah. I will make sure this is incorporated into our API side. cc: @YihuiGuo

xiaoyongzhu:
closed because of https://github.com/linkedin/feathr/pull/258/files

PR #258, opened 2022-05-15
Added feature validation

lhayhurst:
See [this issue](https://github.com/linkedin/feathr/issues/227), "Add validation for feature name".

This PR adds a new method to `feathr_project/feathr/feature.py`, `FeatureBase.validate_feature_name`, that validates the provided feature name against the following rule: 

> Only alphabet, numbers, and '_' are allowed in the name. It can not start with numbers. Note that '.' is NOT ALLOWED

Unit tests are provided for this new function, which is called from the `FeatureBase` constructor before setting the `self.name.`

I drafted a (now closed) PR against my fork, https://github.com/lhayhurst/feathr/pull/1, in order to validate that this PR was headed in the right directi. @hangfei gave me feedback in that PR, which I've incorporated here.


feathr_project/feathr/__init__.py
```diff
@@ -13,3 +13,4 @@
 from .lookup_feature import LookupFeature
 from .aggregation import Aggregation
 from .feathr_configurations import SparkExecutionConfiguration
+from feathr.api.app.core.feathr_api_exception import *
```
lhayhurst, line 16: I made the call that the exceptions in [feathr.api.app.core.fathr_api_exception.py](https://github.com/linkedin/feathr/blob/main/feathr_project/feathr/api/app/core/feathr_api_exception.py) should be exported by default as part of the public API. Happy to roll this back if reviewers feel otherwise.

feathr_project/feathr/api/app/core/feathr_api_exception.py
```diff
@@ -1,8 +1,17 @@
-from app.core.error_code import ErrorCode
-
+from feathr.api.app.core.error_code import ErrorCode
```
lhayhurst, line 1: Looking around the code, it seems that the package root is `feathr`, not `app`, so I corrected the `from app.core.error_code import ErrorCode` while I was here.

feathr_project/feathr/__init__.py
```diff
@@ -13,3 +13,4 @@
 from .lookup_feature import LookupFeature
 from .aggregation import Aggregation
 from .feathr_configurations import SparkExecutionConfiguration
+from .api.app.core.feathr_api_exception import *
```
lhayhurst, line 16: I made the call that the exceptions in [feathr.api.app.core.fathr_api_exception.py](https://github.com/linkedin/feathr/blob/main/feathr_project/feathr/api/app/core/feathr_api_exception.py) should be exported by default as part of the public API. Happy to roll this back if reviewers feel otherwise.

hangfei: This is a good improment. Thanks.

feathr_project/feathr/api/app/core/feathr_api_exception.py
hangfei: @YihuiGuo can comment on this. But as long as your tests passes, it should be ok to commit.

hangfei (approved):
Looks great! Thanks for the contribution!

xiaoyongzhu (approved):

hangfei:
Once the tests passed, we can merge this change.

feathr_project/feathr/api/app/core/feathr_api_exception.py
lhayhurst: thanks -- looking forward to what @YihuiGuo  says here, as it is a little scary to have two different root namespaces floating around in the code!

lhayhurst:
build is failing for reason unrelated to change 

```
   /opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/urllib3/util/retry.py:38: DeprecationWarning: Using 'Retry.DEFAULT_METHOD_WHITELIST' is deprecated and will be removed in v2.0. Use 'Retry.DEFAULT_ALLOWED_METHODS' instead
    warnings.warn(

feathr_project/test/test_config_loading.py: 2 warnings
feathr_project/test/test_feature_materialization.py: 3 warnings
feathr_project/test/test_azure_snowflake_e2e.py: 2 warnings
feathr_project/test/test_azure_spark_e2e.py: 3 warnings
feathr_project/test/test_feature_registry.py: 1 warning
feathr_project/test/test_input_output_sources.py: 2 warnings
feathr_project/test/test_pyduf_preprocessing_e2e.py: 5 warnings
  /opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/databricks_cli/sdk/api_client.py:75: DeprecationWarning: Using 'method_whitelist' with Retry is deprecated and will be removed in v2.0. Use 'allowed_methods' instead
```

hangfei:
The root cause is this failed test: 
```
    def test_feathr_online_store_agg_features():
        """
        Test FeathrClient() get_online_features and batch_get can get feature data correctly.
        """
        test_workspace_dir = Path(__file__).parent.resolve() / "test_user_workspace"
    
        client = snowflake_test_setup(os.path.join(test_workspace_dir, "feathr_config.yaml"))
    
        online_test_table = get_online_test_table_name("snowflakeSampleDemoFeature")
    
        backfill_time = BackfillTime(start=datetime(2020, 5, 20), end=datetime(2020, 5, 20), step=timedelta(days=1))
        redisSink = RedisSink(table_name= online_test_table)
        settings = MaterializationSettings(name="snowflakeSampleDemoFeature",
                                       sinks=[redisSink],
                                       feature_names=['f_snowflake_call_center_division_name',
                                                      'f_snowflake_call_center_zipcode'],
                                       backfill_time=backfill_time)
        client.materialize_features(settings)
        # just assume the job is successful without validating the actual result in Redis. Might need to consolidate
        # this part with the test_feathr_online_store test case
        client.wait_job_to_finish(timeout_sec=900)
    
        res = client.get_online_features(online_test_table, '1',
                                         ['f_snowflake_call_center_division_name', 'f_snowflake_call_center_zipcode'])
    
        assert len(res) == 2
        assert res[0] != None
        assert res[1] != None
        res = client.multi_get_online_features('snowflakeSampleDemoFeature',
                                        ['1', '2'],
                                        ['f_snowflake_call_center_division_name', 'f_snowflake_call_center_zipcode'])
>       assert res['1'][0] != None
E       assert None != None
```

I am re-running the tests to see if it's transistent.

hangfei:
The feature name f_snowflake_call_center_division_name and f_snowflake_call_center_zipcode seems good. It should pass your validation. Doesn't seem to be related to your change.

hangfei:
It seems the failure exist in main branch. Let me do some research to fix it.

feathr_project/feathr/api/app/core/feathr_api_exception.py
blrchen: Those code with namespace starts with app are introduced for feathr rest api which supports feathr ui. I agree have two different root namespaces is causing confusion.

Ideally, we should move api folder out of feathr. The code under api folder does not needs to be included in feathr package as rest api is meant to be deployed as a container app, instead of being consumed as a library. It's okay to feathr rest api have a different root namespace than feathr core as it is a standalone application consumes feathr core. 

I will do the code move when the prototype work for current rest api is completed. In the meanwhile, could you please make sure your code change in feathr core not depends on feathr rest api? Otherwise there will be circular dependency

lhayhurst: Thanks @blrchen!

> In the meanwhile, could you please make sure your code change in feathr core not depends on feathr rest api? Otherwise there will be circular dependency

This a good point. @hangfei asked me to put this new exception into the `api.app.core.error_code` module (see [PR feedback](https://github.com/lhayhurst/feathr/pull/1#discussion_r873210122)). 

I'm not sure what the right thing to do here; this PR is my first sojourn into the feathr code so I don't have good intuition into these sorts of design. As such I have no strong opinions here. Options seem to be:

1) Leave the code as is and risk having a circular dependency 
2) Back out adding the `FeatureNameValidationError` to the  `api.app.core.error_code` module ; back out exporting the exceptions from `feathr/__init__.py`, and simply define the `FeatureNameValidationError` class in `feature.py` and export the `FeatureNameValidationError` from `feathr/__init__.py`. 
3) Some other option? 

Happy to proceed however everyone sees fit! Namespace/package/module design is hard :/ 

Cheers
Lyle

lhayhurst:
> It seems the failure exist in main branch. Let me do some research to fix it.
> 
> 

Sounds good. I will merge main branch into this PR branch once the aim build is green.

xiaoyongzhu:
@lhayhurst thanks for the contribution! I looked thru the test failures and they are unrelated with the changes that you made.  I have a small PR for those (https://github.com/linkedin/feathr/pull/276) in case you are interested in.

Thanks for the contribution!

PR #259, opened 2022-05-16
moved azure ad config to a .env file for better overriding

chinmaytredence:
The routes.tsx file was containing the azure client id and authority hardcoded. This PR has moved these variables into an .env file which could be useful to override with .env.local file while running in local set up. 

.gitignore in main fethr repository was changed to allow .env file to be committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -144,7 +144,7 @@ celerybeat.pid
 *.sage.py
 
 # Environments
-.env
+#.env
 .venv
 env/
 venv/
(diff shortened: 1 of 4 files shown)


blrchen:
Hello @chinmaytredence the pr looks good to me. Could you please not touch code that not needs to change? It seems you might have some style tool which automatically format the ts code? 

This reminds me that style lint might need to be setup on ui project to make code review easiler.

chinmaytredence:
@blrchen thanks for reviewing this PR. Yes, I observed that code formatting while sending the PR. As you rightly said, we have to come up with style lint. I am using prettier for timebeing. 

I can revert back the formatted code. Please suggest.

blrchen (approved):

blrchen:
PR approved. Thanks for your contribution.

xiaoyongzhu (approved):

PR #276, opened 2022-05-18
Fix test failure

xiaoyongzhu:
This PR fixes the test failure in main:
1. Change the Redis table name in one of the test to make sure it's using the right method for online tables
2. extend some test wait time to avoid time out issues


jaymo001 (approved):

xiaoyongzhu:
interestingly not sure why this still triggers the test running with old code, pretty sure I've modified those:
image

hangfei (approved):

PR #282, opened 2022-05-19
Added typecheck for the registry tags

moupriyaroy25:
Small changes to add type check for registry tags to check if registry_tags is a valid Dict[str, str] parameter.
the registry tags are in a dictionary format like this:

{"description":"this is a test feature"},so valid dictionary parameter.
and if the user does something like this

{"description", "this is a test feature"}, it'll also validate it.

### DIFF
--- a/feathr_project/feathr/anchor.py
+++ b/feathr_project/feathr/anchor.py
@@ -30,7 +30,8 @@ class FeatureAnchor(HoconConvertible):
         self.name = name
         self.features = features
         self.source = source
-        self.registry_tags=registry_tags
+        # adding typecheck for registry tags incase user could also do like this {"description", "this is a test feature"} as well as like this dictionary format {"description":"this is a test feature"}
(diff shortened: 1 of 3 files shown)


hangfei (commented):
Thanks for the PR. Could you add a few tests for this?

Also maybe test dicts that contains non-string objects, like {"apple":[list of things]}

moupriyaroy25:
@hangfei Thanks for reviewing. While I am trying to run the tests I am facing some issues during the installation of feathr module on my system. Here is the issue I have mentioned [https://stackoverflow.com/questions/72316535/confluent-kafka-install-error-error-legacy-install-failure](url)

moupriyaroy25:
@hangfei could you please also elaborate briefly about adding test dicts? Do I need to add any test files for this? I am still so new in this case so adding some links to reference docs might help me figure that out.

PR #283, opened 2022-05-19
Enabling Test Suite to run on windows agents on nightly basis

jainr:
The file will run on a schedule

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,173 @@
+name: Nightly Feathr Tests on Windows
+
+on:
+  schedule:
+    - cron: 15 14 * * *
+    
+jobs:
+  sbt_test:
+    runs-on: windows-latest
+    if: github.event_name == 'push' || github.event_name == 'pull_request' || (github.event_name == 'pull_request_target' && contains(github.event.pull_request.labels.*.name, 'safe to test'))
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Set up JDK 8
+        uses: actions/setup-java@v2
+        with:
+          java-version: "8"
+          distribution: "temurin"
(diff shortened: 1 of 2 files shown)


jainr:
This will run the existing CI and E2E tests on windows build machines.

.github/workflows/main.yml
```diff
@@ -0,0 +1,173 @@
...
+          echo "FEATHR_LOCAL_JAR_FULL_NAME_PATH=$(ls target/scala-2.12/*.jar)" >> $GITHUB_ENV
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Install Feathr Package
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install pytest pytest-xdist databricks-cli
+          python -m pip install -e ./feathr_project/
```
blrchen, line 71: >python -m pip install -e ./feathr_project/ [](http://example.com/codeflow?start=10&length=42)

This might not work on windows Environment. The kafa package won't install.

.github/workflows/main.yml
```diff
@@ -0,0 +1,173 @@
+name: Nightly Feathr Tests on Windows
+
+on:
+  schedule:
+    - cron: 15 14 * * *
```
blrchen, line 5: Better add a comment here explaining what cron expression is resolved to.

jainr: Good comment, will do.

jainr: @blrchen  @xiaoyongzhu DO you know why this file won't show up under GH actions workflows? 
image

.github/workflows/main.yml
jainr: And what is the workaround for that?

blrchen: I haven't found any workaround so far

PR #288, opened 2022-05-20
Fix millisecond timestamp handling

jaymo001:
The current SWA code does not support timestamp with millisecond, e.g. 2019-05-03T15:00:00.123Z


hangfei:
Could you explain the bug in the description?

hangfei (approved):

xiaoyongzhu (approved):

xiaoyongzhu:
we don't have to add `safe-to-test` label now if it's not a forked repo now.

Issue #263, opened 2022-05-16
get_features_from_registry not getting all the features


Manjunath07:
We are doing a PoC to test registering features, getting features from Purview and get offline feature with the Feathr Client for NYX Taxi fare data

Following are the different notebooks to define different features, registering and trying to get features registered in Purview 

**Notebook1:**
1)Project_name=feathrpoc
2)Initiate Feathrclient
3)Feature declaration ['f_location_avg_fare', 'f_trip_time_duration','f_trip_time_rounded']
4)Client build feature
5)Client register features
6)Client list features 

![image](https://user-images.githubusercontent.com/26295302/168567100-2c803055-f911-4d1f-a2de-6e08c7c7a633.png)

**Notebook2:**
1)Project_name=feathrpoc
2)Initiate Feathrclient
3)Feature declaration ['f_location_max_fare'](with UDF)
4)Client build feature
5)Client register features
6)Client list features ( I am able to see the features registered via Notebook1
![image](https://user-images.githubusercontent.com/26295302/168567689-ec379793-bfdc-44b7-af62-5c5bb14a4c42.png)

**Notebook3:**
1)Project_name=feathrpoc
2)Initiate Feathrclient
3)Client list features
4)Client get feature from registry
![image](https://user-images.githubusercontent.com/26295302/168568294-e223daad-89fc-4993-9b69-35eb594b3fa2.png)

With above(Notebook3) it is only getting 2 features from registry('f_location_max_fare','f_trip_time_rounded') and not able to get other features present in registry('f_location_avg_fare', 'f_trip_time_duration')

xiaoyongzhu:
interesting, it shouldn't be. @Manjunath07 assuming you are using the latest version (0.4.0)?

Manjunath07:
@xiaoyongzhu  latest feathr from https://pypi.org/project/feathr/  doesn't have get_features_from_registry function so I have cloned the main branch and tried installing via setup.py and used the function

xiaoyongzhu:
@YihuiGuo will be working on this issue

YihuiGuo:
@Manjunath07  Could you share more details of your notebook? for example , how did you programmatically perform the "feature declaration"  , local feature building and remote registration  for three features in Notebook #1 ?

xiaoyongzhu:
@YihuiGuo is able to repro the issue and figured out the root cause. The fix should be available soon.

YihuiGuo:
[...]

PR #304, opened 2022-05-30
Yihui/moderate registration conflict

YihuiGuo:
This PR merges existing anchor with new anchor during registration in order to fix concurrent conflict.
Will serve as a quick fix, but will not solved it entirely. 
Full fix will work with MVCC, and is in progress.


YihuiGuo:
Related issue: https://github.com/linkedin/feathr/issues/263

feathr_project/test/test_feature_registry.py
```diff
@@ -62,7 +64,33 @@ def test_feathr_register_features_e2e():
                                 output_path=output_path)
     client.wait_job_to_finish(timeout_sec=900)
 
+def test_feathr_register_features_partially():
```
jaymo001, line 67: Could you elaborate the use case in the PR description? And it would be helpful to explain the exact meaning of "register features partially".

Yuqing-cat (approved):
Looks good to me as a short-term fix.

feathr_project/test/test_feature_registry.py
YihuiGuo: Updated the PR description. Comments has been added to this test function.

PR #303, opened 2022-05-27
Feathr UI: API spec alignment and ux experience improvement

blrchen:
This PR update ui code to align with latest api spec which abstracts common interfaces among different registration providers. This PR also enhances lineage ux experience and adds RBAC experience for flighting.

Staging env: https://brave-moss-045d2ce00-6.eastasia.1.azurestaticapps.net/

## New Feature registration API integration
API spec can be viewed at https://feathr-registry.azurewebsites.net/docs.  UI code is now purview free.

## UX improvements
- [x] When a feature is clicked, the feature node is highlighted
- [x] When a feature is clicked, the feature level lineage path is highlighted. Feature level lineage means direct producer and consumer.
- [x] Node in graph are colored grouped by feature type.
- [x] User can toggle graph to filter by feature type.
- [x] Flow arrow is enabled to indicate producer and consumer relationship
- [x] All out of scope resources are removed in graph, for example, Altas Process, Project.
- [x] Add navigate button in node so user can click to navigate to feature detail page.
- [x] RBAC in flgihting, accessible via /management path. 

Linked issue: #272

PR #311, opened 2022-06-01
Python SQL Registry

windoze:
This is the SQL-based Feathr implemented in Python.
The API spec is updated to include some previously missing pieces.
The corresponding UI changes is in another PR #303  submitted by @blrchen 

There are still some TODOs:
1. User still needs to manually create the database schema by running the script `script/schema.sql`.
2. The docker file is included, but there is not ARM template to deploy by one click.

We can talk about these tasks later, right now, please focus on the implementation itself.


feathr_project/registry/registry/database.py
```diff
@@ -0,0 +1,84 @@
...
+def quote(id):
+    if isinstance(id, str):
+        return f"'{id}'"
+    else:
+        return ",".join([f"'{i}'" for i in id])
+
+
+def parse_conn_str(s: str) -> dict:
+    """
+    TODO: Not a sound and safe implementation
```
xiaoyongzhu, line 24: Put TODO in comments? In the method description, we should say the connection string format that we are expecting (guess it should be standard connection string and maybe we can put some reference)

feathr_project/registry/registry/database.py
```diff
@@ -0,0 +1,84 @@
...
+        "user": parts["User ID"],
+        "password": parts["Password"],
+        # "charset": "utf-8",   ## For unknown reason this causes connection failure
+    }
+
+
+class MssqlConnection(DbConnection):
+    @staticmethod
+    def connect(*args, **kwargs):
+        conn_str = os.environ["CONNECTION_STR"]
```
xiaoyongzhu, line 41: update docs somewhere so end users know which environment to set. Also `CONNECTION_STR` sounds like a general one, maybe we should call it something specific to Feathr?

feathr_project/registry/requirements.txt
```diff
@@ -0,0 +1,3 @@
+pymssql
+fastapi
+uvicorn
```
xiaoyongzhu, line 3: How is this file used?

feathr_project/registry/main.py
```diff
@@ -0,0 +1,77 @@
...
+from registry.models import EntityType
+
+rp = "/"
+try:
+    rp = os.environ["API_BASE"]
+    if rp[0] != '/':
+        rp = '/' + rp
+except:
+    pass
+print("Using API BASE: ", rp)
```
xiaoyongzhu, line 16: Use logger rather than print?

feathr_project/registry/registry/interface.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @abstractmethod
+    def search_entity(self,
+                      keyword: str,
+                      type: list[EntityType],
+                      project: Optional[Union[str, UUID]] = None) -> list[EntityRef]:
+        pass
+
+
+if __name__ == '__main__':
```
xiaoyongzhu, line 45: remove this part

feathr_project/registry/registry/database.py
windoze: It's just ADO connection string, I'm just not sure if the implementation covered all cases, it works if you simply copy the connection string from the Azure portal, but not thoroughly tested for all other cases.
This value is not from the external input so there is no security issue.

feathr_project/registry/registry/database.py
windoze: It's just the generic format, not specific to Feathr.
This configuration is set in different ways for different deployment methods, e.g. AppService for Container and AKS have very different configuration mechanisms to deploy containers/pods.
@jainr is preparing the user manual and that should cover this part.

feathr_project/registry/requirements.txt
windoze: User can run `pip install -r requirements.txt` if they want to start from the source code, also this file is used by Docker image building process.
This is standard Python working process so no doc needed, if user decided to work with the source code they should already know.

feathr_project/registry/main.py
windoze: Agreed.
This program doesn't have enough logs, I plan to add them.
Do we need it in this PR or later?

feathr_project/registry/registry/interface.py
windoze: Removed

feathr_project/registry/registry/models.py
```diff
@@ -0,0 +1,707 @@
+from abc import ABC, abstractmethod
+from enum import Enum
+from typing import Optional, Union
+from uuid import UUID
+import json
+import re
+
+
+def to_snake(d):
```
xiaoyongzhu: Add description for this method. Also we should add hint for the input/output types.

feathr_project/registry/registry/models.py
```diff
@@ -0,0 +1,707 @@
...
+    if isinstance(d, str):
+        return re.sub('([A-Z]\w+$)', '_\\1', d).lower()
+    if isinstance(d, list):
+        return [to_snake(i) if isinstance(i, (dict, list)) else i for i in d]
+    return {to_snake(a): to_snake(b) if isinstance(b, (dict, list)) else b for a, b in d.items()}
+
+
+def to_type(value, type):
```
xiaoyongzhu: Add description for this method. Also we should add hint for the input/output types.

feathr_project/registry/registry/models.py
```diff
@@ -0,0 +1,707 @@
...
+                pass
+        return type[value]
+    return type(value)
+
+
+def to_uuid(value):
+    return to_type(value, UUID)
+
+
+class ValueType(Enum):
```
xiaoyongzhu: Add description for this method.

feathr_project/registry/registry/models.py
```diff
@@ -0,0 +1,707 @@
...
+    TENSOR = 0
+
+
+class TensorCategory(Enum):
+    DENSE = 0
+    SPARSE = 1
+
+
+class EntityType(Enum):
+    Project = 1
```
xiaoyongzhu: Add description for this method.

xiaoyongzhu:
I have a few comments in general:
1. for folder names etc. we should reflect it's a "SQL" based registry
2. We should add docs so users know how to use this. Should they build docker files? And how do they connect to the docker file? At least I'm not clear on those.
3. Seems like many code are missing comments, which will be hard for other devs to maintain. Please do add comments/descriptions for all the functions.

feathr_project/registry/registry/models.py
```diff
@@ -0,0 +1,707 @@
...
+        self.transform_expr = transform_expr
+
+    def to_dict(self) -> dict:
+        return {
+            "transform_expr": self.transform_expr
+        }
+
+
+class WindowAggregationTransformation(Transformation):
+    def __init__(self,
```
xiaoyongzhu: We should avoid duplicated feature definition code. Maybe extend the existing class?

xiaoyongzhu:
for those DerivedFeatureDef or AnchorFeatureDef classes, can we rename it so that we know it's specific to registry? Currently it's a bit confusing as there might be duplications on the existing feature definition that is defined here, vs the ones that are available in the "Feathr Core" code.

windoze:
> 

The Dockerfile is not intent to be used by users directly, it is used by the building process, and we'll setup CI actions to publish ready-to-use docker images to some public repo like DockerHub, @jainr is working on this task.

xiaoyongzhu:
> > 
> 
> The Dockerfile is not intent to be used by users directly, it is used by the building process, and we'll setup CI actions to publish ready-to-use docker images to some public repo like DockerHub, @jainr is working on this task.

How would the end users use that pre-built container in this case?

windoze:
> 

That's why I suggested to split the registry from the feathr client repo.
They share a lot of concepts hence lots of similar but not exactly same classes and types, they're similar because they're representing the same thing, not same because they're for different purposes, on the client side, they're for user composing and job submitting, on the server side, they're for displaying, storing and indexing.
If we put these to separated components together, the confusion is inevitable.

windoze:
> > > 
> > 
> > 
> > The Dockerfile is not intent to be used by users directly, it is used by the building process, and we'll setup CI actions to publish ready-to-use docker images to some public repo like DockerHub, @jainr is working on this task.
> 
> How would the end users use that pre-built container in this case?

Again, this part is covered by the manual that @jainr is currently working on, so let's put the comments about the deploying/running questions in that upcoming user-manual PR.

feathr_project/registry/registry/models.py
windoze: The type hint is omitted purposely, this function is for dynamically type casting, the input value could be anything and the output value is in the type specified by `type` parameter.
This is not something can be handled by Python typing mechanism.

xiaoyongzhu:
> > 
> 
> That's why I suggested to split the registry from the feathr client repo. They share a lot of concepts hence lots of similar but not exactly same classes and types, they're similar because they're representing the same thing, not same because they're for different purposes, on the client side, they're for user composing and job submitting, on the server side, they're for displaying, storing and indexing. If we put these to separated components together, the confusion is inevitable.

Correct me if I'm wrong here, but seems like the registry in this PR is only to be consumed by UI (and the REST API layer). I.e. it's not ready to be consumed by the Python client (say, users cannot call `get_features_from_registry()` for this registry.

For this specific PR, it should be self contained (that's why you have a dockerfile) and it's equivalent of building another smaller Purview service, I think we should move it under a folder *NOT* in a separate repo, but rather in the same repo but not in the python client folder. maybe create something under root folder and put it there. Or put it under the `ui/` folder.

Also - this folder (https://github.com/linkedin/feathr/tree/main/feathr_project/feathr/api) should be also moved tin a similar way (i.e. not using registry service)

windoze:
> > > 
> > 
> > 
> > That's why I suggested to split the registry from the feathr client repo. They share a lot of concepts hence lots of similar but not exactly same classes and types, they're similar because they're representing the same thing, not same because they're for different purposes, on the client side, they're for user composing and job submitting, on the server side, they're for displaying, storing and indexing. If we put these to separated components together, the confusion is inevitable.
> 
> Correct me if I'm wrong here, but seems like the registry in this PR is only to be consumed by UI (and the REST API layer). I.e. it's not ready to be consumed by the Python client (say, users cannot call `get_features_from_registry()` for this registry.
> 
> For this specific PR, it should be self contained (that's why you have a dockerfile) and it's equivalent of building another smaller Purview service, I think we should move it under a folder _NOT_ in a separate repo, but rather in the same repo but not in the python client folder. maybe create something under root folder and put it there. Or put it under the `ui/` folder.
> 
> Also - this folder (https://github.com/linkedin/feathr/tree/main/feathr_project/feathr/api) should be also moved tin a similar way (i.e. not using registry service)


Agree.

Please raise an issue and let's make decision on directory structure and do a major restructuring soon, not only for this part, but also other components.

xiaoyongzhu:
> > > > 
> > > 
> > > 
> > > That's why I suggested to split the registry from the feathr client repo. They share a lot of concepts hence lots of similar but not exactly same classes and types, they're similar because they're representing the same thing, not same because they're for different purposes, on the client side, they're for user composing and job submitting, on the server side, they're for displaying, storing and indexing. If we put these to separated components together, the confusion is inevitable.
> > 
> > 
> > Correct me if I'm wrong here, but seems like the registry in this PR is only to be consumed by UI (and the REST API layer). I.e. it's not ready to be consumed by the Python client (say, users cannot call `get_features_from_registry()` for this registry.
> > For this specific PR, it should be self contained (that's why you have a dockerfile) and it's equivalent of building another smaller Purview service, I think we should move it under a folder _NOT_ in a separate repo, but rather in the same repo but not in the python client folder. maybe create something under root folder and put it there. Or put it under the `ui/` folder.
> > Also - this folder (https://github.com/linkedin/feathr/tree/main/feathr_project/feathr/api) should be also moved tin a similar way (i.e. not using registry service)
> 
> Agree.
> 
> Please raise an issue and let's make decision on directory structure and do a major restructuring soon, not only for this part, but also other components.



> > > > 
> > > 
> > > 
> > > That's why I suggested to split the registry from the feathr client repo. They share a lot of concepts hence lots of similar but not exactly same classes and types, they're similar because they're representing the same thing, not same because they're for different purposes, on the client side, they're for user composing and job submitting, on the server side, they're for displaying, storing and indexing. If we put these to separated components together, the confusion is inevitable.
> > 
> > 
> > Correct me if I'm wrong here, but seems like the registry in this PR is only to be consumed by UI (and the REST API layer). I.e. it's not ready to be consumed by the Python client (say, users cannot call `get_features_from_registry()` for this registry.
> > For this specific PR, it should be self contained (that's why you have a dockerfile) and it's equivalent of building another smaller Purview service, I think we should move it under a folder _NOT_ in a separate repo, but rather in the same repo but not in the python client folder. maybe create something under root folder and put it there. Or put it under the `ui/` folder.
> > Also - this folder (https://github.com/linkedin/feathr/tree/main/feathr_project/feathr/api) should be also moved tin a similar way (i.e. not using registry service)
> 
> Agree.
> 
> Please raise an issue and let's make decision on directory structure and do a major restructuring soon, not only for this part, but also other components.

Can we do two things here?
1. Move this implementation to the project root or somewhere else, rather than under /feathr_project
2. Add a bit on the implementation notes for how this solution works in the README file (like we are only storing edges and nodes, and each nodes have different type, etc.)

xiaoyongzhu:
> > > > 
> > > 
> > > 
> > > That's why I suggested to split the registry from the feathr client repo. They share a lot of concepts hence lots of similar but not exactly same classes and types, they're similar because they're representing the same thing, not same because they're for different purposes, on the client side, they're for user composing and job submitting, on the server side, they're for displaying, storing and indexing. If we put these to separated components together, the confusion is inevitable.
> > 
> > 
> > Correct me if I'm wrong here, but seems like the registry in this PR is only to be consumed by UI (and the REST API layer). I.e. it's not ready to be consumed by the Python client (say, users cannot call `get_features_from_registry()` for this registry.
> > For this specific PR, it should be self contained (that's why you have a dockerfile) and it's equivalent of building another smaller Purview service, I think we should move it under a folder _NOT_ in a separate repo, but rather in the same repo but not in the python client folder. maybe create something under root folder and put it there. Or put it under the `ui/` folder.
> > Also - this folder (https://github.com/linkedin/feathr/tree/main/feathr_project/feathr/api) should be also moved tin a similar way (i.e. not using registry service)
> 
> Agree.
> 
> Please raise an issue and let's make decision on directory structure and do a major restructuring soon, not only for this part, but also other components.



> > > > 
> > > 
> > > 
> > > That's why I suggested to split the registry from the feathr client repo. They share a lot of concepts hence lots of similar but not exactly same classes and types, they're similar because they're representing the same thing, not same because they're for different purposes, on the client side, they're for user composing and job submitting, on the server side, they're for displaying, storing and indexing. If we put these to separated components together, the confusion is inevitable.
> > 
> > 
> > Correct me if I'm wrong here, but seems like the registry in this PR is only to be consumed by UI (and the REST API layer). I.e. it's not ready to be consumed by the Python client (say, users cannot call `get_features_from_registry()` for this registry.
> > For this specific PR, it should be self contained (that's why you have a dockerfile) and it's equivalent of building another smaller Purview service, I think we should move it under a folder _NOT_ in a separate repo, but rather in the same repo but not in the python client folder. maybe create something under root folder and put it there. Or put it under the `ui/` folder.
> > Also - this folder (https://github.com/linkedin/feathr/tree/main/feathr_project/feathr/api) should be also moved tin a similar way (i.e. not using registry service)
> 
> Agree.
> 
> Please raise an issue and let's make decision on directory structure and do a major restructuring soon, not only for this part, but also other components.

Can we do two things here?
1. Move this implementation to the project root or somewhere else, rather than under /feathr_project
2. Add a bit on the implementation notes for how this solution works in the README file (like we are only storing edges and nodes, and each nodes have different type, etc.)

xiaoyongzhu:
Can we do a few things here?

1. Move this implementation to the project root or somewhere else, rather than under /feathr_project
2. Add a bit on the implementation notes for how this solution works in the README file (like we are only storing edges and nodes, and each nodes have different type, etc.)
3. Add comments in the python files (especially the ones under `registry` folder) to make it easier to maintain

xiaoyongzhu:
Can we do a few things here?

1. Move this implementation to the project root or somewhere else, rather than under /feathr_project
2. Add a bit on the implementation notes for how this solution works in the README file (like we are only storing edges and nodes, and each nodes have different type, etc.)
3. Add comments in the python files (especially the ones under `registry` folder) to make it easier to maintain

blrchen (approved):

Yuqing-cat (approved):

PR #316, opened 2022-06-02
Improve readability for sample notebooks

xiaoyongzhu:
This PR improves the notebook readability by:
- [x] use JPG instead of PNG to make the graph readable in dark theme
- [x] reduce the content on provisioning resources to make the notebook more focused.


blrchen (approved):

hangfei:
These two are duplicated:

hangfei:
These 2 needs to be rewritten, they seem to be simliar(at least in title):

hangfei:
There are four prerequisites. We probably should reduce them or rename it.

From a higher-level, i feel we should move this long boil-plate to somewhere else so it's easier to navigate the notebook. Right now, it's gigantic. But i also understand it's hard to do it right now. Hopefully we can find a way in the future.

Issue #326, opened 2022-06-03
Azure job url is not set and thus not logged correctly during error.


bozhonghu:
It seems like the job url for the azure job is not set. The error log here does not work: https://github.com/linkedin/feathr/blob/85c2bf50337bdd32f3e63bb521d9aa55f994736c/feathr_project/feathr/_synapse_submission.py#L138
```
logger.error("Feathr job has failed. Please visit this page to view error message: {}", self.job_url)
E               AttributeError: '_FeathrSynapseJobLauncher' object has no attribute 'job_url'
feathr_project/feathr/_synapse_submission.py:138: AttributeError
```
^ is the error I see. 
It seems like the variable self.job_url is not set anywhere

windoze:
Fixed in #327

PR #327, opened 2022-06-04
#326 Azure Synapse client doesn't have job_url

windoze:
#326
Azure Synapse client doesn't have `job_url`, remove it from the error log.


xiaoyongzhu (approved):

xiaoyongzhu:
The failure should be irrelevant of this fix.

hangfei (approved):
Approved the PR. Thanks for the fix.

Curious is there a way to get the job_url?

xiaoyongzhu:
> Approved the PR. Thanks for the fix.
> 
> Curious is there a way to get the job_url?

I guess no... For synapse it's a bit painful as it requires the subscription id which we don't have in the job, but @windoze might be able to comment as well

PR #333, opened 2022-06-08
Optimize environment variable behavior

xiaoyongzhu:
Currently it's a bit confusing for end users and there are multiple asks on what are all the user configurable environment variables (and the expected behaviors).

Also, we only read a portion of the env variables in key vault, but we should actually support reading all the env variables in KV.

This PR solves the above issues, by having a clear logic on how to get the env variables in the following order:

1. If the key is set in the envrionment variable, Feathr will use the value of that environment variable
2. If it's not set in the environment, then a value is retrieved from from the feathr_config.yaml file with the same config key.
3. If it's not available in the feathr_config.yaml file, Feathr will try to reterive the value from a key vault service. Currently only Azure Key Vault is supported.

Also add docs describing all the env variables.


feathr_project/feathr/secrets/akv_client.py
```diff
@@ -15,8 +15,10 @@ def get_akv_secret(self, secret_name: str):
                 credential=DefaultAzureCredential()
             )
         try:
+            # replace '_' with '-' since Azure Key Vault doesn't support it
```
Yuqing-cat, line 18: Shall we split it into a separate function, say get_akv_name_from_config_name(): for better maintenance?
Also, we could add a note in docs to make sure that key vault name and config name always follow the same naming rules, e.g., in https://github.com/linkedin/feathr/blob/main/docs/how-to-guides/azure-deployment.md

docs/how-to-guides/feathr-configuration-and-env.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+This allows end users to store the configuration in a secure way, say in Kubernetes secrets, key vault, etc. All the configurations available for end users to configure are listed below.
+
+
+# Default behaviors
+
+Feathr will get the configurations in the following order:
+
+1. If the key is set in the envrionment variable, Feathr will use the value of that environment variable
```
blrchen, line 30: >envrionment [](http://example.com/codeflow?start=28&length=11)

envrionment => environment

docs/how-to-guides/feathr-configuration-and-env.md
```diff
@@ -0,0 +1,114 @@
+---
+layout: default
+title: Configuration, environment varialbles, and store secrets in a secure way
```
blrchen, line 3: varialbles => variables 

Same applies to line 7

docs/how-to-guides/feathr-configuration-and-env.md
```diff
@@ -0,0 +1,114 @@
...
+
+This allows end users to store the configuration in a secure way, say in Kubernetes secrets, key vault, etc. All the configurations available for end users to configure are listed below.
+
+
+# Default behaviors
+
+Feathr will get the configurations in the following order:
+
+1. If the key is set in the envrionment variable, Feathr will use the value of that environment variable
+2. If it's not set in the environment, then a value is retrieved from from the feathr_config.yaml file with the same config key.
```
blrchen, line 31: >a value is retrieved from from the feathr_config.yaml file [](http://example.com/codeflow?start=44&length=58)

retrieved from from the feathr_config.yaml file

Remove duplicated from

feathr_project/feathr/utils/_envvariableutil.py
```diff
@@ -24,51 +29,59 @@ def get_environment_variable_with_default(self, *args):
...
-                # resolve one layer after another
-                for arg in args:
-                    yaml_layer = yaml_layer[arg]
-                return yaml_layer
-            except KeyError as exc:
-                logger.info(exc)
+                return  self.akv_client.get_feathr_akv_secret(env_keyword) 
+            except ResourceNotFoundError:
+                # print out warning message if cannot find the env variable in all the resources
+                logger.warning('Environment variable {} not found in environment variable, default YAML config file, or key vault service.', env_keyword)
```
blrchen, line 61: >warning [](http://example.com/codeflow?start=23&length=7)

For required enivronment vars, if the value is not found here, shoulg log error instead and stop running

feathr_project/feathr/utils/_envvariableutil.py
```diff
@@ -24,51 +29,59 @@ def get_environment_variable_with_default(self, *args):
...
-                        ' is not set in the environment variables.')
-            akv_name = os.environ.get("KEY_VAULT_NAME")
-            if akv_name:
-                logger.info('Fetching the value {} from Key Vault.', variable_key)
-                akv_client = AzureKeyVaultClient(akv_name)
-                password = akv_client.get_akv_secret(variable_key)
-        return password
 
+        if password:
+            return password
```
blrchen, line 80: >password [](http://example.com/codeflow?start=19&length=8)

get_environment_variable is a generic method and use password as value name here is confused.

docs/how-to-guides/feathr-configuration-and-env.md
```diff
@@ -0,0 +1,114 @@
...
+3. If it's not available in the feathr_config.yaml file, Feathr will try to reterive the value from a key vault service. Currently only Azure Key Vault is supported.
+
+# A list of environment variables that Feathr uses
+| Environment Variable     | Description     |
```
blrchen, line 35: How about move require status to a new column, the value can be Required or Required if uses XXX. Then description can only focus on indicate what environment var is used for.

docs/how-to-guides/feathr-configuration-and-env.md
```diff
@@ -0,0 +1,114 @@
+---
```
hangfei, line 1: Which file is this going to replace? Seems another setup wiki.

xiaoyongzhu: yes this is a separate doc/wiki.

docs/how-to-guides/feathr-configuration-and-env.md
xiaoyongzhu: Thanks, updated

feathr_project/feathr/utils/_envvariableutil.py
xiaoyongzhu: renamed to `env_var_value`

feathr_project/feathr/utils/_envvariableutil.py
xiaoyongzhu: Ideally we should do so, but currently we load all the variables even if it's optional. This is worth a separate PR to solve though.

docs/how-to-guides/feathr-configuration-and-env.md
```diff
@@ -0,0 +1,114 @@
...
+layout: default
+title: Configuration, environment variables, and store secrets in a secure way
```
blrchen, line 3: Suggest add a link to this doc in https://github.com/linkedin/feathr/blob/main/feathr_project/feathrcli/data/feathr_user_workspace/feathr_config.yaml

xiaoyongzhu: good point, will add a link to that once we merge this to main.

blrchen (approved):

hangfei (approved):

Issue #211, opened 2022-05-03
Consider using Maven based spark job submission


xiaoyongzhu:
Since Feathr is on Maven now (https://search.maven.org/artifact/com.linkedin.feathr/feathr_2.12), we should consider using Maven as the source to submit Spark jobs,rather than using the public wasb path, since it's a bit hard to maintain and doesn't distribute well (like don't have mirrors etc., and will be slow).

Databricks supports maven based library: https://docs.microsoft.com/en-us/azure/databricks/dev-tools/api/latest/libraries#--library

Synapse can be achieved by using this config: `spark.jars.packages`. See (https://spark.apache.org/docs/3.2.1/configuration.html) for more details.

PR #334, opened 2022-06-09
Windoze/211 maven submission

windoze:
#211 Fix, enable submitting Feathr JAR from Maven.


feathr_project/feathr/spark_provider/_synapse_submission.py
```diff
@@ -92,21 +97,45 @@ def submit_feathr_job(self, job_name: str, main_jar_path: str = None,  main_clas
...
+                current_dir = pathlib.Path(__file__).parent.resolve()
+                main_jar_path = os.path.join(current_dir, "noop-1.0.jar")
+            else:
+                # This is a PySpark job
+                pass
+        main_jar_cloud_path = None
+        if main_jar_path:
+            # Now we have a main jar, either feathr or noop
+            if main_jar_path.startswith('abfs'):
+                main_jar_cloud_path = main_jar_path
```
xiaoyongzhu, line 130: nit: if main_jar_path and main_jar_path.startswith('abfs'):  ?

xiaoyongzhu:
Overall look good to me. Seems like the test failure is irrelevant.

xiaoyongzhu:
Can we also add a bit test case to cover this?

xiaoyongzhu (approved):

xiaoyongzhu:
Pasting some background info for this PR: 
```
Since Feathr is on Maven now (https://search.maven.org/artifact/com.linkedin.feathr/feathr_2.12), we should consider using Maven as the source to submit Spark jobs,rather than using the public wasb path, since it's a bit hard to maintain and doesn't distribute well (like don't have mirrors etc., and will be slow).
```

jainr:
+1 to adding tests to cover this. Looks good to me.

windoze:
E2E test was added in the latest commit.

PR #337, opened 2022-06-09
enable JWT token param in frontend API calls

Yuqing-cat:
- Enable JWT token to replace current hardcode mock token.
- Protect all API calls with JWT token
- Abstract MsalInstance logic into `utils.tsx`

### DIFF
--- a/ui/src/api/api.tsx
+++ b/ui/src/api/api.tsx
@@ -161,8 +161,8 @@ export const getIdToken = async( msalInstance: PublicClientApplication ): Promis
       scopes: ["User.Read"],
       account: activeAccount || accounts[0]
   };
+  // Silently acquire an token for a given set of scopes. Will use cached token if available, otherwise will attempt to acquire a new token from the network via refresh token.
   await msalInstance.acquireTokenSilent(request).then(response => {
-    console.log(response.idToken)
     return response.idToken
   }).catch(error => {
      if (error instanceof InteractionRequiredAuthError) {
--- a/ui/src/api/api.tsx
+++ b/ui/src/api/api.tsx
@@ -9,7 +9,7 @@ const MOCK_TOKEN = "mockAppServiceKey";
 const msalInstance = getMsalConfig();
 
 export const fetchDataSources = async (project: string) => {
-  const token = await getAccessToken(msalInstance);
+  const token = await getIdToken(msalInstance);
   return Axios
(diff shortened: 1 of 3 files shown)


ui/src/api/api.tsx
```diff
@@ -1,11 +1,15 @@
 import Axios from "axios";
 import { DataSource, Feature, FeatureLineage, UserRole, Role } from "../models/model";
+import { InteractionRequiredAuthError, PublicClientApplication } from "@azure/msal-browser";
 import mockUserRole from "./mock/userrole.json";
+import { getMsalConfig } from "../utils/utils";
 
 const API_ENDPOINT = process.env.REACT_APP_API_ENDPOINT + "/api/v1";
-const token = "mockAppServiceKey";
+const MOCK_TOKEN = "mockAppServiceKey";
```
blrchen, line 8: const MOCK_TOKEN = "mockAppServiceKey";
this should get removed

ui/src/api/api.tsx
```diff
@@ -139,3 +153,23 @@ export const deleteUserRole = async (role: Role) => {
...
 }
+
+export const getIdToken = async( msalInstance: PublicClientApplication ): Promise<string> => {
+  const activeAccount = msalInstance.getActiveAccount(); // This will only return a non-null value if you have logic somewhere else that calls the setActiveAccount API
+  const accounts = msalInstance.getAllAccounts();
+  const request = {
+      scopes: ["User.Read"],
+      account: activeAccount || accounts[0]
+  };
+  await msalInstance.acquireTokenSilent(request).then(response => {
```
blrchen, line 164: >acquireTokenSilent [](http://example.com/codeflow?start=21&length=18)

It seems acquireTokenSilent will look up token from local cache first, if not found or expired, this method will fetch / refresh token. Is this correct?

If refresh token is taken be MADL internally, it's better add a comment here.

ui/src/api/api.tsx
```diff
@@ -139,3 +153,23 @@ export const deleteUserRole = async (role: Role) => {
...
+    console.log(response.idToken)
```
blrchen, line 165: >console.log(response.idToken) [](http://example.com/codeflow?start=4&length=29)

console.log(response.idToken)
Let's remove this

blrchen (approved):
:shipit:

xiaoyongzhu (approved):

PR #352, opened 2022-06-13
Adding webapp and webappPlan creation to deployment ARM template 

jainr:
- Adding support to create a linux app service plan
- Adding support to create an Azure webapp for containers that uses that plan and hosts a public dockerhub image
- Fixing a bug on instructions to fetch principalId, objectId is replaced by id property
https://docs.microsoft.com/en-us/cli/azure/microsoft-graph-migration#breaking-changes

Successful deployment
image

Also adding instructions to create AAD app and to get clientId and TenantId.
Made corresponding changes in tempplate to accept them as parameters and pass it to webapp as settings.


docs/how-to-guides/azure_resource_provision.json
```diff
@@ -13,7 +13,7 @@
         "principalId": {
             "type": "String",
             "metadata": {
-                "description": "Specifies the principal ID assigned to the role. You can find it by logging into 'https://shell.azure.com/bash' and run 'az ad signed-in-user show --query objectId -o tsv'"
```
Yuqing-cat, line 16: Create a separate PR to fix all occurrence of this. Please take a look.
https://github.com/linkedin/feathr/pull/360

Yuqing-cat (commented):
Shall we have a "provisionWebApp" parameter to let user decide whether to deploy it or not just like Purview & EventHub?

docs/how-to-guides/azure_resource_provision.json
jainr: We always want to provision the webapp, so its not an optional feature since that would mean we are not provisioning UI or API server

Yuqing-cat (approved):

xiaoyongzhu:
title says it's still WIP so not sure it's ready to take review or not?

But anyway I looked thru the changes and I'm good with those.

docs/how-to-guides/azure_resource_provision.json
```diff
@@ -349,6 +353,48 @@
...
+            "kind": "app,linux,container",
+            "identity": {
+                "type": "systemAssigned"
+            },
+            "properties": {
+                "name": "[variables('webAppName')]",
+                "httpsOnly": "true",
+                "serverFarmId": "[resourceId('Microsoft.Web/serverfarms', variables('webAppPlanName'))]",
+                "siteConfig": {
+                    "linuxFxVersion": "DOCKER|dockercloud/hello-world",
```
blrchen, line 387: For integration let's use this docker image name blrchen/feathr-sql-registry

Should move this to parameter so it can be overrided. As currently sql registry and purview registry are different images. Furthermore, in the further, we might use a different name spaces under org account.

jainr: Fixed

edwincheung (approved):

docs/how-to-guides/azure_resource_provision.json
```diff
@@ -18,28 +18,47 @@
...
-      "allowedValues": ["true", "false"],
+      "allowedValues": [ "true", "false" ],
       "type": "String",
       "metadata": {
         "description": "Whether or not to deploy eventhub provision script"
       }
-    }
+    },
+    "dockerImage": {
+      "defaultValue": "blrchen/feathr-sql-registry",
```
Yuqing-cat, line 44: Are we going to have Blair's alias in the default value?

docs/how-to-guides/azure-deployment.md
```diff
@@ -15,10 +15,35 @@ Feathr has native cloud integration and getting started with Feathr is very stra
...
+
+# Updating the SPA app created above, currently there is no CLI support to add redirectUris to a SPA, so we have to patch manually via az rest
+az rest --method PATCH --uri "https://graph.microsoft.com/v1.0/applications/$aad_objectId" --headers "Content-Type=application/json" --body "{spa:{redirectUris:['https://$sitename.azurewebsites.net/.auth/login/aad/callback']}}"
+
+# Make a note of the ClientId and TenantId, you will need it during deployment.
+echo "AAD_CLIENT_ID: $aad_clientId"
+echo "AZURE_TENANT_ID: $aad_tenantId"
+``` 
+
+3. Click the button below to deploy a minimal set of Feathr resources. This is not for production use as we choose a minimal set of resources, but treat it as a template that you can modify for further use. Note that you should have "Owner" access in your subscription to perform some of the actions.
```
Yuqing-cat, line 46: Actually, user can choose Purview Registry or SQL Registry, and for access control it will also be configurable. Thus, currently deployment seems not `minimal` per my understanding.

Yuqing-cat (approved):

docs/how-to-guides/azure_resource_provision.json
jainr: Blair is going to change it and then I will update as part of a different PR

docs/how-to-guides/azure-deployment.md
jainr: I did not add that part, but agreed we should change it to be more relevant.

jainr: Let me address this too in another PR with container registry changes.

xiaoyongzhu:
image
I looked thru the new experience, as above, a few feedbacks:
1. Docker Image - seems the description is wrong there. Also we should avoid using a specific person's name (like `blchen`)
2. Maybe rename "Docker Image" as "Pre-built Feathr UI Docker Image" or something like that so users know what we are talking about
3. We should probably combine the object ID and this AAD credential together since anyway they need to run thru the scripts

jainr:
Thanks for the feedback @xiaoyongzhu, I will include that in subsequent PR

PR #360, opened 2022-06-14
obejectId query cmd update

Yuqing-cat:
If user is using Azure CLI Version <= 2.36, then the original objectId works. 
However, the cloud shell always run the latest version. Thus, I just simply update command in each occurrence. 

(Please check the issue link to find more backgrounds about Azure CLI upgrade)

Also, I'd like to raise a potential risk of doc maintenance: this command exists in 8 files and 5 of them are notebook / markdown guide. Will it be better to abstract the content into a single "How-to-provision-azure-resource" and link to it?

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -33,7 +33,7 @@ Feathr automatically computes your feature values and joins them to your trainin
 
 Feathr has native cloud integration. To use Feathr on Azure, you only need three steps:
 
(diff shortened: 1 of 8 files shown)


jainr (approved):
Thanks for making this fix with objectId, I also made the fix for this in my ARM template PR but in only covers the template.

I am not sure if you reformatted but there seems a big delta in ARM template, was any other thing changed except objectId?

jainr:
I also agree with consolidating this in a single place, please share your ideas.

PR #365, opened 2022-06-14
Fix test failures on main

xiaoyongzhu:
Currently the main branch has test failures, fixing those by:
1. running `pytest` only on `feathr_project/test/` folder
2. fixing a bug in `get_environment_variable` method.

### DIFF
--- a/.github/workflows/pull_request_push_test.yml
+++ b/.github/workflows/pull_request_push_test.yml
@@ -116,7 +116,7 @@ jobs:
 
         run: |
           # run only test with databricks. run in 4 parallel jobs
-          pytest  -n 8 feathr_project/test/
+          pytest  -n 12 feathr_project/test/
 
   azure_synapse_test:
       # might be a bit duplication to setup both the azure_synapse test and databricks test, but for now we will keep those to accelerate the test speed
@@ -183,6 +183,6 @@ jobs:
         run: |
           # skip databricks related test as we just ran the test; also seperate databricks and synapse test to make sure there's no write conflict
           # run in 4 parallel jobs to make the time shorter
-          pytest  -n 8 feathr_project/test/
+          pytest  -n 12 feathr_project/test/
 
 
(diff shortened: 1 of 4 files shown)


.github/workflows/pull_request_push_test.yml
```diff
@@ -116,7 +116,7 @@ jobs:
 
         run: |
           # run only test with databricks. run in 4 parallel jobs
-          pytest -n 4
+          pytest  -n 6 feathr_project/test/
```
hangfei, line 119: extra whitespace

.github/workflows/pull_request_push_test.yml
```diff
@@ -183,6 +183,6 @@ jobs:
         run: |
           # skip databricks related test as we just ran the test; also seperate databricks and synapse test to make sure there's no write conflict
           # run in 4 parallel jobs to make the time shorter
-          pytest -n 4
+          pytest  -n 6 feathr_project/test/
```
hangfei, line 186: extra whitespace

xiaoyongzhu: fixed

hangfei (approved):

feathr_project/test/test_input_output_sources.py
```diff
@@ -36,7 +38,7 @@ def test_feathr_get_offline_features_with_parquet():
     now = datetime.now()
     # set output folder based on different runtime
     if client.spark_runtime == 'databricks':
-        output_path = ''.join(['dbfs:/feathrazure_cijob','_', str(now.minute), '_', str(now.second), ".parquet"])
+        output_path = ''.join(['dbfs:/feathrazure_cijob','_', str(now.minute), '_', str(now.second),'_', str(now.microsecond),  ".parquet"])
```
blrchen, line 41: Is there any specifc reason only add this for databricks

xiaoyongzhu: In very rare cases databricks will fail to read the result file (though the spark job is successful). Assuming it might be related with the output folder write conflict. This might apply to Synapse as well but I never see that in Synapse so didn't add it there.

blrchen (approved):

PR #368, opened 2022-06-15
Use elegant way to get all features for a project and fix potential concurrency issues

xiaoyongzhu:
This PR addresses a few comments/issues:

1. Previously, when getting all the features for a certain project, we use `search_entities` method, which might not be scalable when there are tons of entities (or might take some time)
2. We use the attribute field to get the relations between different entities (between project and anchor, or between derived features, etc.), which might be hard to merge two projects.

This PR solves those two issues, by doing the following:

1. Using `lineage` API to get all the entities belonging to a certain project at once, so that we don't have to do search, rather we rely on the pre-computed purview lineage to get all the entities
2. when calling `get_features_from_registry()`, we don't use the attributes since there might be concurrency issues; Instead, we use `AtlasProcess` to avoid potential concurrency issues.

Also fixed an issue brought by https://github.com/linkedin/feathr/pull/380 where additional schema is added.


hangfei (approved):

xiaoyongzhu:
https://github.com/linkedin/feathr/pull/466 the issues should already be fixed by this PR hence closing.

Issue #371, opened 2022-06-15
Purview-based API doesn not work with UI


YihuiGuo:
Currently Feathr UI is implemented based on SQL backend. With Purview backend however, some data formats and interfaces does not align, so for now if we run Feathr UI with Purview backend, it won't work. 

Need to align these backends, specifically, change purview api data format and interfaces , to become identical to SQL version

blrchen:
Closing as the major code for support purview as registry provider is completed and integrated with Web UI.

A demo [site](https://aka.ms/feathrdemo-purview]) is setup for hosting latest Purview based registry code.

PR #372, opened 2022-06-15
Yihui/align purview api with UI

YihuiGuo:
Currently Feathr UI is implemented based on SQL backend. With Purview backend however, some data formats and interfaces does not align, so for now if we run Feathr UI with Purview backend, it won't work.

Need to align these backends, specifically, change purview api data format and interfaces , to become identical to SQL version

Issue: https://github.com/linkedin/feathr/issues/371


feathr_project/feathr/api/app/core/configs.py
```diff
@@ -1,12 +1,12 @@
...
 import os
 
-API_Version = "/v1"
+API_Version = "/api/v1"
 log_level = os.getenv("logLevel", "INFO")
 app_insights_connection_string = os.getenv("APPLICATIONINSIGHTS_CONNECTION_STRING")
 formatter = logging.Formatter("[%(asctime)s] [%(name)s:%(lineno)s - %(funcName)5s()] %(levelname)s - %(message)s")
 appServiceKey =  "mockAppServiceKey" # os.getenv("AppServiceKey")
-
+default_purview_name = "feathrazuretest3-purview1"
```
jainr, line 9: Why are we hardcoding the name here?

YihuiGuo: I will work with @blrchen on adding this to env, and read from env in later updates, thanks for reminding

feathr_project/feathr/registry/_feature_registry_purview.py
```diff
@@ -936,9 +903,9 @@ def _list_registered_entities_with_details(self, project_name: str, entity_type:
         # append the guid list. Since we are using project_name + delimiter to search, all the result will be valid.
         guid_list = [entity["id"] for entity in result_entities]
 
-        entity_res = [] if guid_list is None or len(guid_list)==0 else self.purview_client.get_entity(
-            guid=guid_list)["entities"]
-        return entity_res
+        guid_batches = [guid_list[i:i + limit] for i in range(starting_offset, len(guid_list), limit)]
```
xiaoyongzhu, line 906: add a bit comment here saying it's using paging?

feathr_project/feathr/api/app/backends/purview_backend/purview_backend.py
```diff
@@ -27,6 +27,10 @@ def ListProjects(self):
         raw_result = registry_client._list_registered_entities_with_details(entity_type=[TYPEDEF_FEATHR_PROJECT])
```
xiaoyongzhu, line 27: Shall we also change the `self.default_project_name = "feathr_github_ci_synapse"` to make it configurable?

xiaoyongzhu:
Overall LGTM. Thanks for the PR!

PR #380, opened 2022-06-17
Refreshed JDBC support

windoze:
This is the refreshed PR for #124, the old one was stale for too long and cannot be cleanly merged.
This PR does following things:

1. Add a `--system-properties` command line parameter to the Spark job, so the client can pass multiple secrets to the job.
2. The reading from JDBC function already exists in the job code but doesn't have public interface to use, now this function is enabled when you have JdbcSource in the feature definition file.
3. On the client side, a `JdbcSource` is added to represent a data source from JDBC, with auth, and corresponding data model change.
4. Various fixes to both job and client sides.
5. An E2E test to get offline features from JDBC data source.

Currently the E2E test cannot run in the CI env because it requires a database with `green_tripdata_2020-04.csv` imported into a table called `green_tripdata_2020_04`, with auth configured correctly.

To run the E2E test locally, you need:

- Create a SQL database instance with user/pass auth on Azure, AAD/token auth is not covered by this test.
- Import `green_tripdata_2020-04.csv` into the database with the table name set to `green_tripdata_2020_04`, be aware of the column type, all numeric columns like prices or distances must be "FLOAT" and all ID columns must be "INT" if you're using AzureSQL.
- Set following environment variables
    - `nycTaxiBatchJdbcSource_USER`: The user name to login to database server
    - `nycTaxiBatchJdbcSource_PASSWORD`: The password to login to database server

Then you should be able to run the `test_sql_source.py` directly or via `pytest`, both Databricks and Azure Synapse should work. You can use whatever user/pass combination for each Jdbc data source, not limited to the previously global unique `JDBC_*` settings.

### DIFF
--- a/feathr_project/feathr/registry/_feature_registry_purview.py
+++ b/feathr_project/feathr/registry/_feature_registry_purview.py
@@ -35,7 +35,7 @@ from feathr.constants import *
 from feathr.definition.feature import Feature, FeatureType,FeatureBase
 from feathr.definition.feature_derivations import DerivedFeature
 from feathr.definition.repo_definitions import RepoDefinitions
-from feathr.definition.source import HdfsSource, InputContext, Source
+from feathr.definition.source import HdfsSource, InputContext, JdbcSource, Source
 from feathr.definition.transformation import (ExpressionTransformation, Transformation,
                                    WindowAggTransformation)
 from feathr.definition.typed_key import TypedKey
@@ -93,6 +93,14 @@ class _FeatureRegistry(FeathrRegistry):
 
                 AtlasAttributeDef(
                     name="path", typeName="string", cardinality=Cardinality.SINGLE),
+                AtlasAttributeDef(
+                    name="url", typeName="string", cardinality=Cardinality.SINGLE),
+                AtlasAttributeDef(
+                    name="dbtable", typeName="string", cardinality=Cardinality.SINGLE),
+                AtlasAttributeDef(
+                    name="query", typeName="string", cardinality=Cardinality.SINGLE),
+                AtlasAttributeDef(
+                    name="auth", typeName="string", cardinality=Cardinality.SINGLE),
                 AtlasAttributeDef(name="event_timestamp_column",
                                   typeName="string", cardinality=Cardinality.SINGLE),
                 AtlasAttributeDef(name="timestamp_format",
@@ -289,7 +297,7 @@ class _FeatureRegistry(FeathrRegistry):
                 transformed_original_elements.setdefault(elem['qualifiedName'],elem)
             return list(transformed_original_elements.values())
             
-    def _parse_source(self, source: Union[Source, HdfsSource]) -> AtlasEntity:
+    def _parse_source(self, source: Union[Source, HdfsSource, JdbcSource]) -> AtlasEntity:
         """
         parse the input sources
         """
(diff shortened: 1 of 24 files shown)


src/main/scala/com/linkedin/feathr/offline/source/dataloader/BatchDataLoader.scala
```diff
@@ -63,16 +63,14 @@ private[offline] class BatchDataLoader(ss: SparkSession, location: InputLocation
     val dataIOParametersWithSplitSize = Map(SparkIOUtils.SPLIT_SIZE -> inputSplitSize) ++ dataIOParameters
     val dataPath = location.getPath
 
-    log.info(s"Loading ${dataPath} as DataFrame, using parameters ${dataIOParametersWithSplitSize}")
+    log.info(s"Loading ${location} as DataFrame, using parameters ${dataIOParametersWithSplitSize}")
     try {
-      if (dataPath.startsWith("jdbc")){
-        JdbcUtils.loadDataFrame(ss, dataPath)
-      } else {
```
windoze, line 70: @blee1234 May I ask why did you write this? InputLocation.loadDf exists since the beginning, why do you hack all the way to break its function?

src/main/scala/com/linkedin/feathr/offline/transformation/AnchorToDataSourceMapper.scala
```diff
@@ -93,13 +94,19 @@ private[offline] class AnchorToDataSourceMapper(dataPathHandlers: List[DataPathH
       failOnMissingPartition: Boolean): DataFrame = {
 
     val dataLoaderHandlers: List[DataLoaderHandler] = dataPathHandlers.map(_.dataLoaderHandler)
-    val pathChecker = PathChecker(ss)
-    val pathAnalyzer = new TimeBasedHdfsPathAnalyzer(pathChecker, dataLoaderHandlers)
```
windoze, line 97: @blee1234 still, this line does **not** work with not only Jdbc source but also Kafka, 
Please **DO NOT** assume that every data source is on HDFS, it is **NOT TRUE**.

edwincheung: @windoze Suggest to add comment in the code to explain this intention. Will be easier for folks to understand.

windoze: > 

I did add comments in my code, this message is my comment on original code, seems he forgot that a data source could be something other than HDFS even before this PR.

src/main/scala/com/linkedin/feathr/offline/source/dataloader/BatchDataLoader.scala
blee1234: hey @windoze I actually just refactored that path from location.getPath to dataPath. 

https://github.com/linkedin/feathr/commit/3dc354969a6bc470eeae52276910cfa0fe09130d#diff-28e361a441ef37dd98fad74ea933606d35c3ad7a1beb5bf0c026438fdd8b42f3R68-R70

I don't think that refactoring is what you're referencing in your comment. The whole segment you're referencing was actually in the repo from this commit.

https://github.com/linkedin/feathr/commit/3b1a2ea089afba724ff8dba14f10b913e161149a#diff-8d2002a450fb2814a262118070cf8a9ed0564b76c27a0cd7c4360cf6480c8e5fR64-R66

where the file was originally called SparkDataLoader and then renamed later to BatchDataLoader.

src/main/scala/com/linkedin/feathr/offline/transformation/AnchorToDataSourceMapper.scala
blee1234: @windoze sorry I'm confused on your comment. for line 97, my change simply passes dataLoaderHandlers as a parameter to the the TimeBasedHdfsPathAnalyzer. 

If you're looking at who added the TimeBasedHdfsPathAnalyzer in the first place, you'll have to go through the git history.

xiaoyongzhu:
`--system-properties` sounds very generic and might be confusing for future developers, and assuming the intent is only to pass the credentials, maybe we can call something closer to `credentials`? Like `--jdbc-credentials` ?, or `--client-secrets`?

feathr_project/feathr/registry/_feature_registry_purview.py
```diff
@@ -93,6 +93,14 @@ def _register_feathr_feature_types(self):
 
                 AtlasAttributeDef(
                     name="path", typeName="string", cardinality=Cardinality.SINGLE),
+                AtlasAttributeDef(
+                    name="url", typeName="string", cardinality=Cardinality.SINGLE),
+                AtlasAttributeDef(
+                    name="dbtable", typeName="string", cardinality=Cardinality.SINGLE),
```
xiaoyongzhu, line 99: Can we add those in a customized dict or something so it's more flexible, rather than adding it here in the attributes? Reasons being:
1. Those schemas are hard (or impossible) to change
2. Not everyone is using JDBC
Putting those in a `map<string,string>` allows us to have more flexibility.

src/main/scala/com/linkedin/feathr/offline/transformation/AnchorToDataSourceMapper.scala
windoze: > 

Sorry for looking at the wrong guy, I was just wondering, why the code is fully of assumptions that the data source comes solely from HDFS and makes even `local override` doesn't fully work as expected.
I'll start a major cleanup soon, and wish you can provide some help.

feathr_project/feathr/registry/_feature_registry_purview.py
windoze: I don't plan to do larger refactor to this part in this PR, yihui is working on the data model re-design and will touch this as well. So let keep this change as simple as possible, to reduce the probability to have conflict with her change.

BTW all these fields are optional so it won't break existing HDFS support.

windoze: > 

As to `--system-properties`, if you look into the code on Scala side, you'll find that the name does serve the purpose.

Yuqing-cat (approved):

blrchen (approved):

PR #198, opened 2022-04-30
Code format with the right subfolders

xiaoyongzhu:
Goal of this PR is to add the right folder hierarchy for Feathr code (so that it can be better organized). 

Other small improvements include:
- Updated the pydoc structure to make sure all the pydocs are in the right module name. 
- Renaming frameconfig to feathrconfig and put it into the right folder structure
- Docs on how to expose the right module in pydocs

pydocs in this PR: https://feathr-xiaoyzhu.readthedocs.io/en/latest/index.html
pydocs in main branch: https://feathr.readthedocs.io/

We should also move the `api` folder out from the python client. Waiting for @blrchen and @YihuiGuo to fix those as there seems to be some dependencies. It's described at https://github.com/linkedin/feathr/issues/287.

This should have no end user impact as end users are using this:
`from feathr import X`

The new file structure is like this:

image

```
(base) xiazhu@xiazhu-mn1 feathr % tree feathr_project/feathr -I __pycache__
feathr_project/feathr
├── __init__.py
├── api
│   ├── app
│   │   ├── __init__.py
│   │   ├── backends
│   │   │   ├── backend_base.py
│   │   │   └── purview_backend
│   │   │       ├── __init__.py
│   │   │       ├── purview_backend.py
│   │   │       └── purview_backend_config.py
│   │   ├── core
│   │   │   ├── __init__.py
│   │   │   ├── configs.py
│   │   │   ├── error_code.py
│   │   │   ├── error_handling.py
│   │   │   ├── feathr_api_exception.py
│   │   │   └── registry.py
│   │   ├── main.py
│   │   ├── models
│   │   │   └── features.py
│   │   └── routes
│   │       ├── feature.py
│   │       ├── job.py
│   │       ├── offline_feature.py
│   │       ├── online_feature.py
│   │       ├── project.py
│   │       └── transformation.py
│   └── requirements.txt
├── client.py
├── constants.py
├── definition
│   ├── _materialization_utils.py
│   ├── aggregation.py
│   ├── anchor.py
│   ├── dtype.py
│   ├── feathrconfig.py
│   ├── feature.py
│   ├── feature_derivations.py
│   ├── lookup_feature.py
│   ├── materialization_settings.py
│   ├── query_feature_list.py
│   ├── repo_definitions.py
│   ├── settings.py
│   ├── sink.py
│   ├── source.py
│   ├── transformation.py
│   └── typed_key.py
├── exception
├── protobuf
│   ├── __init__.py
│   └── featureValue_pb2.py
[...]

PR #383, opened 2022-06-20
Fix Python doc links in Docs and add `HdfsSink` back

Yuqing-cat:
Fix the side effect of https://github.com/linkedin/feathr/pull/198
- Update all Python doc links in docs
- Add `HdfsSink` into `__all__ ` section of `feathr_project/feathr/__init__.py`
- Add notes in `docs/dev_guide/update_python_docs.md`

Note:
- As mentioned in `docs/dev_guide/update_python_docs.md`:
    - Add an `__all__` section in `__init__.py` (see code [here](https://github.com/linkedin/feathr/blob/main/feathr_project/feathr/__init__.py)). Every components that are included in the `__all__` section is exposed to end users. Others are not exposed in the pydocs.

I'm not sure if the `HdfsSink` is not included by design. If it shouldn't expose to users, then the section in `docs/concepts/feature-generation.md` should be removed.

### DIFF
--- a/feathr_project/feathr/__init__.py
+++ b/feathr_project/feathr/__init__.py
@@ -45,6 +45,7 @@ __all__ = [
     'MaterializationSettings',
     'MonitoringSettings',
     'RedisSink',
+    'HdfsSink',
     'MonitoringSqlSink',
     'FeatureQuery',
     'LookupFeature',
--- a/docs/concepts/feature-definition.md
+++ b/docs/concepts/feature-definition.md
@@ -28,14 +28,14 @@ batch_source = HdfsSource(name="nycTaxiBatchSource",
                           timestamp_format="yyyy-MM-dd HH:mm:ss")
 ```
 
-See the [Python API documentation](https://feathr.readthedocs.io/en/latest/feathr.html#feathr.source.HdfsSource) to get the details on each input column.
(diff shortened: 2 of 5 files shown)


jainr (approved):

xiaoyongzhu (approved):

xiaoyongzhu:
Thanks! In order to make it a bit more user friendly (i.e. users can import from the `feathr` module, it's a bit manual to add user facing APIs there.

There were some automated ways, but seems they are too automated and need many manual changes to tell whether a certain API is needed or not.

Yuqing-cat:
> Thanks! In order to make it a bit more user friendly (i.e. users can import from the `feathr` module, it's a bit manual to add user facing APIs there.
> 
> There were some automated ways, but seems they are too automated and need many manual changes to tell whether a certain API is needed or not.

I see. 
For this PR, I only covered all Python API links in Docs. Currently, we don't have an official guide for which modules should be exposed to users or not. Thus, I cannot help to check if there are other missing modules in __all__.

feathr_project/feathr/__init__.py
```diff
@@ -45,6 +45,7 @@
     'MaterializationSettings',
     'MonitoringSettings',
     'RedisSink',
+    'HdfsSink',
```
blrchen, line 48: i am curious why materialize test can pass without this., does that means HdfsSink already included in other packages?

PR #385, opened 2022-06-22
Feathr UI: remove delete button

blrchen:
This PR removes delete button on both feature list and feature detail pages.


jainr:
Why are we removing the delete button?

blrchen:
> Why are we removing the delete button?

Remove as user case is not clear and showing this in UI causes many confusions during review.

Yuqing-cat (approved):

xiaoyongzhu (approved):

PR #387, opened 2022-06-23
Entity creation APIs

windoze:
SQL registry now support entity creation.
Before this PR the SQL registry solely worked with UI without creating functions.


YihuiGuo (approved):

registry/sql-registry/registry/models.py
```diff
@@ -276,6 +276,7 @@ def to_dict(self) -> dict:
 class Attributes(ToDict):
     @staticmethod
     def new(entity_type: Union[str, EntityType], **kwargs):
+        print("YYY ", entity_type, kwargs)
```
Yuqing-cat, line 279: What's the usage of this print?

Yuqing-cat (commented):
This is a very important prerequisite PR for access control features. Very helpful enhancement.
I would suggest moving `database.py` to a common folder under `/registry` so that sql-based access control can reuse it. 
Also, are the `interface.py` and `models.py` a common description for all backend registries? If so, shall we move those common definition outside of `sql-registry`?

registry/sql-registry/registry/database.py
```diff
@@ -53,8 +58,11 @@ def __init__(self, params):
     def make_connection(self):
         self.conn = pymssql.connect(**self.params)
 
-    def execute(self, sql: str, *args, **kwargs) -> list[dict]:
-        debug(f"SQL: `{sql}`")
+    def query(self, sql: str, *args, **kwargs) -> list[dict]:
+        """
+        Make SQL query and return result
+        """
+        warn(f"SQL: `{sql}`")
```
blrchen, line 65: >warn(f"SQL: `{sql}`") [](http://example.com/codeflow?start=8&length=21)

Not sure if it's intented to print every raw sql.

If it's intented, probably should log as verbose instead of warning

registry/sql-registry/registry/db_registry.py
```diff
@@ -88,7 +93,237 @@ def get_project(self, id_or_name: Union[str, UUID]) -> EntitiesAndRelations:
...
+                        and attr.key == definition.key:
+                    # Creating exactly same entity
+                    # Just return the existing id
+                    return _to_uuid(r[0]["entity_id"])
+                raise ValueError("Entity %s already exists" %
+                                 definition.qualified_name)
+            r1 = []
+            if definition.input_anchor_features:
+                c.execute(
+                    fr'''select entity_id, entity_type, qualified_name from entities where entity_id in ({quote(definition.input_anchor_features)}) and entity_type = 'feathr_anchor_feature_v1' ''')
```
blrchen, line 282: >feathr_anchor_feature_v1 [](http://example.com/codeflow?start=167&length=24)

should make these const enum

registry/sql-registry/registry/db_registry.py
```diff
@@ -139,22 +375,24 @@ def _get_entity(self, id_or_name: Union[str, UUID]) -> Entity:
...
+        if not ids:
+            return []
+        rows = self.conn.query(fr'''select entity_id, qualified_name, entity_type, attributes
             from entities
             where entity_id in ({quote(ids)})
         ''')
         ret = []
         for row in rows:
             row["attributes"] = json.loads(row["attributes"])
+            print("XXX", row)
```
blrchen, line 387: >print [](http://example.com/codeflow?start=12&length=5)

pls clean up print

registry/sql-registry/main.py
```diff
@@ -74,4 +78,39 @@ def get_feature_lineage(feature: str) -> dict:
...
 
-app.include_router(prefix = rp, router=router)
+@router.post("/projects")
+def new_project(definition: dict) -> UUID:
+    id = registry.create_project(ProjectDef(**to_snake(definition)))
+    return {"guid": str(id)}
+
+
+@router.post("/projects/{project}/datasources")
+def new_project_datasource(project: str, definition: dict) -> UUID:
```
blrchen, line 88: >UUID [](http://example.com/codeflow?start=62&length=4)

def new_project_datasource(project: str, definition: dict) -> UUID:
    project_id = registry.get_entity_id(project)
    id = registry.create_project_datasource(project_id, SourceDef(**to_snake(definition)))
    return {"guid": str(id)}

I guess you meant return json, not UUID? Same applies to other post methods

registry/sql-registry/registry/models.py
```diff
@@ -546,15 +547,20 @@
         return self._input_anchor_features + self._input_derived_features
 
     @input_features.setter
-    def input_features(self, v: Union[dict, Entity]):
+    def input_features(self, v: Union[dict, Entity, EntityRef]):
```
blrchen, line 550: please change e, f, v to a meaningful name to make this code easiler to read

registry/sql-registry/test/test_create.py
```diff
@@ -0,0 +1,44 @@
...
+project1 = r.get_entity(project1_id)
+assert project1.qualified_name == "unit_test_project_1"
+
+# Re-create project, should return the same id
+id = r.create_project(ProjectDef("unit_test_project_1"))
+assert project1_id == id
+
+source1_id = r.create_project_datasource(project1_id, SourceDef(
+    qualified_name="unit_test_project_1__source1", name="source1", path="hdfs://somewhere", type="hdfs"))
+print("source1 id ", source1_id)
```
blrchen, line 28: >print [](http://example.com/codeflow?start=0&length=5)

print("source1 id ", source1_id)

Should asset result instead of print result, same apply to the rest of code

registry/sql-registry/registry/models.py
windoze: Removed.

registry/sql-registry/registry/database.py
windoze: Log level changed to debug

registry/sql-registry/test/test_create.py
windoze: assertions added

registry/sql-registry/main.py
windoze: Fixed

registry/sql-registry/registry/db_registry.py
windoze: Changed to use Enums

windoze:
> This is a very important prerequisite PR for access control features. Very helpful enhancement. I would suggest moving `database.py` to a common folder under `/registry` so that sql-based access control can reuse it. Also, are the `interface.py` and `models.py` a common description for all backend registries? If so, shall we move those common definition outside of `sql-registry`?

I plan to submit another following up PR for re-structuring, don't want to mess up before function stabilization.

Yuqing-cat (approved):

Yuqing-cat:
> plan to submit another following up PR for re-structuring, don't want to mess up before function stabilization.

Agree, I can also help to do this my access control PR.

registry/sql-registry/main.py
```diff
@@ -64,7 +67,8 @@ def get_project_features(project: str, keyword: Optional[str] = None) -> list:
 def get_feature(feature: str) -> dict:
```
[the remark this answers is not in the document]
windoze: Fixed

registry/sql-registry/registry/models.py
```diff
@@ -432,9 +465,9 @@ def __init__(self,
         self._source = None
```
YihuiGuo, line 465: Source and features are used for relation, but shall we add source id param to create_project_anchor_feature? 
otherwise there is no way for anchor features to see "source" of the anchor they belong to.

windoze: The source id is set elsewhere.
XYZAttributes classes are **not** supposed to be created manually, it's a part of `Entity` object and should be deserialized from JSON/dict or converted from XYZDef, former is to read from database, latter is to handle user creation requests.
If you want to create AnchorAttributes, first you should have a AnchorDef object, which is provided by user in POST req, then call its `to_attr` method to create the blanked AnchorAttributes, I cannot fill `source` here because there is only an UUID in AnchorDef, but in AnchorAttributes we need a `EntityRef` which also includes the qualified name, this name can only be found when you have the whole registry data. So you need to manually fetch this source Entity object and set it as the property `source` of AnchorAttributes object.
Because we don't support updating right now, so the `_features` property is always empty when an anchor just got created, you can just leave it as is.

Yuqing-cat (approved):

PR #397, opened 2022-06-25
Replace PurView client with Registry API client

windoze:
We'll no longer support directly accessing PurView, the feature registration process must to go through API layer, so I removed the PurView client and replace the implementation with a Registry API client.

A specific test suite is added to make sure the serialization/deserialization works and the client can communicate with registry service correctly.

The new implementation maintains the same interface as the old one except the `__init__`, so I made necessary change to the code where it was used.


feathr_project/test/test_registry_client.py
```diff
@@ -0,0 +1,307 @@
...
+    assert len(anchors)==2
+    request_features = [a for a in anchors if a.name=='request_features'][0]
+    assert isinstance(request_features.source, InputContext)
+    assert len(request_features.features)==4
+    aggregationFeatures = [a for a in anchors if a.name=='aggregationFeatures'][0]
+    assert len(aggregationFeatures.features)==2
+    assert isinstance(aggregationFeatures.source, HdfsSource)
+    assert aggregationFeatures.source.name=="nycTaxiBatchSource"
+    assert aggregationFeatures.source.path=="wasbs://public@azurefeathrstorage.blob.core.windows.net/sample_data/green_tripdata_2020-04.csv"
+    assert len(derived_features)==3
```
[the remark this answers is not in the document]
windoze: This project is **not** what we created in `test_create`, it's a predefined one in the test db.

You can check out there for more details.

feathr_project/test/test_user_workspace/feathr_config.yaml
```diff
@@ -102,6 +102,7 @@ online_store:
     ssl_enabled: True
 
 feature_registry:
+  endpoint: "https://feathr-sql-registry.azurewebsites.net/api/v1"
```
xiaoyongzhu, line 105: Also add comments on this (maybe in the doc here https://github.com/linkedin/feathr/blob/main/docs/how-to-guides/feathr-configuration-and-env.md) so end users know what's this about

feathr_project/feathr/registry/_feathr_registry_client.py
```diff
@@ -0,0 +1,630 @@
...
+    ret._qualified_name = v["attributes"]["qualifiedName"]
+    ret._registry_id = UUID(v["guid"])
+    return ret
+
+
+def _get_type_name(v: Any) -> str:
+    if isinstance(v, Source):
+        return "feathr_source_v1"
+    elif isinstance(v, FeatureAnchor):
+        return "feathr_anchor_v1"
```
xiaoyongzhu, line 540: use `TYPE_DEF_ANCHOR` etc. instead of hard coded string

feathr_project/feathr/registry/_feathr_registry_client.py
```diff
@@ -0,0 +1,630 @@
...
+    if not r.ok:
+        raise RuntimeError(
+            f"Failed to call registry API, status is {r.status_code}, error is {r.text}")
+    return r
+
+
+def source_to_def(v: Source) -> dict:
+    ret = {}
+    if v.name == "PASSTHROUGH":
+        return {
```
xiaoyongzhu, line 335: avoid using hadcode string?

xiaoyongzhu:
From design point of view, I think it's good to reroute the registry requests to the REST API, which should be pluggable as that also means customer has an additional component and they might need to maintain that. Also there are some existing customers who call purview directly and we need a more friendly way to migrate them.

I suggest we do the following:
- have two classes, one is the existing `_FeatureRegistry` and another is the updated one (maybe `_FeatureRegistryWithRESTAPI` or something like that (this is actually a bad name and we might want a more friendly class name)
- if `endpoint` is configured, then call `_FeatureRegistryWithRESTAPI`, otherwise if it's not configured, call `_FeatureRegistry`
- might be some code duplications so we can either keep two copy of the code, or unify those common functions into one place

So that it will be easier for existing users to use the existing solutions and make the architecture simple, while if users want to use RBAC or other backends, they can simply add `endpoint`  to configure

I left a few minor comments in the code.

xiaoyongzhu:
Also please add comments for the newly added methods so it's easier to maintain

feathr_project/test/test_user_workspace/feathr_config.yaml
```diff
@@ -102,6 +102,7 @@ online_store:
...
   purview:
```
xiaoyongzhu, line 106: nit: `rest_endpoint` to make it a bit more reflective?

Also I'm not sure if the purview section is needed or not if we are using REST API, given that those purview account info will be maintained in the REST API layer?

windoze:
> 

I don't think it's a good idea.

The new PurView backed registry has updated data model which is not fully compatible with the original one, if we keep both clients, there is a good chance that users use them in parallel and corrupt whole data set without cure.

We should only preserve one PurView client, either old one, or new one through registry backend, if we don't want to cause trouble to both our users and ourselves.

feathr_project/test/test_user_workspace/feathr_config.yaml
windoze: My suggestion is, since we're going to remove PurView client, we should also remove the following `purview` section.
In this case `feature_registry.endpoint` won't cause any confusion.

feathr_project/feathr/registry/_feathr_registry_client.py
windoze: LOL I was looking for this constant in the source code but forgot its exact name, thanks for pointing it out.

feathr_project/feathr/registry/_feathr_registry_client.py
windoze: Fixed

feathr_project/test/test_user_workspace/feathr_config.yaml
windoze: Renamed to `api_endpoint`

feathr_project/feathr/registry/_feathr_registry_client.py
```diff
@@ -0,0 +1,631 @@
...
+    def __init__(self, project_name: str, endpoint: str, project_tags: Dict[str, str] = None, credential=None, config_path=None):
+        self.project_name = project_name
+        self.project_tags = project_tags
+        self.endpoint = endpoint
+        self.credential = DefaultAzureCredential(
+            exclude_interactive_browser_credential=False) if credential is None else credential
+        self.project_id = None
+
+    def register_features(self, workspace_path: Optional[Path] = None, from_context: bool = True, anchor_list=[], derived_feature_list=[]):
+        """Register Features for the specified workspace.
```
xiaoyongzhu, line 51: Specify type hint in the method parameters for `anchor_list` and `derived_feature_list`

feathr_project/feathr/registry/_feathr_registry_client.py
```diff
@@ -0,0 +1,631 @@
...
+    def _create_project(self) -> UUID:
+        r = self._post(f"/projects", {"name": self.project_name})
+        self.project_id = UUID(r["guid"])
+        return self.project_id
+
+    def _create_source(self, s: Source) -> UUID:
+        r = self._post(
+            f"/projects/{self.project_id}/datasources", source_to_def(s))
+        id = UUID(r["guid"])
+        s._registry_id = id
```
xiaoyongzhu, line 109: Would a `Source` class have a `_registry_id` attribute? I don't see it in the `Source` definition?

feathr_project/feathr/registry/_feathr_registry_client.py
```diff
@@ -0,0 +1,631 @@
...
+"""
+        )
+        derived_feature_configs = tm.render(
+            derived_features=repo_definitions.derived_features)
+        config_file_path = os.path.join(local_workspace_dir, config_file_name)
+        write_to_file(content=derived_feature_configs,
+                      full_file_name=config_file_path)
+
+
+def check(r):
```
xiaoyongzhu, line 326: add type hint for `r`

should this be `_check` given it's internal?

feathr_project/feathr/registry/_feathr_registry_client.py
```diff
@@ -0,0 +1,631 @@
...
+        ret["eventTimestampColumn"] = v.event_timestamp_column
+    if v.timestamp_format:
+        ret["timestampFormat"] = v.timestamp_format
+    if v.registry_tags:
+        ret["tags"] = v.registry_tags
+    return ret
+
+
+def dict_to_source(v: dict) -> Source:
+    id = UUID(v["guid"])
```
xiaoyongzhu, line 373: Should this be an internal method so should start with _? Also add a bit comments for the purpose of this method.

feathr_project/feathr/registry/_feathr_registry_client.py
```diff
@@ -0,0 +1,631 @@
...
+    return ret
+
+
+def dict_to_anchor(v: dict) -> FeatureAnchor:
+    ret = FeatureAnchor(name=v["attributes"]["name"],
+                        source=None,
+                        features=[],
+                        registry_tags=v["attributes"].get("tags", {}),
+                        __no_validate=True)
+    ret._source_id = UUID(v["attributes"]["source"]["guid"])
```
xiaoyongzhu, line 425: same q as above - is `_source_id` part of `FeatureAnchor`?

feathr_project/feathr/registry/_feathr_registry_client.py
windoze: All these `_***_id` properties are temporary during the `register_features` process.

Entities have relations, and the relations are represented as connections between GUIDs on the registry server side, so I need to record registry_id for each entity to build correct connections between them, otherwise I have to retrieve GUID from the server side by their qualified names one by one, which could be very slow in some cases.

So all these `_***_id` are just for performance boosting solely, they have no use outside of this scope so I don't think we need to add them into the public interface.

blrchen (approved):

windoze:
There are some comments unresolved, I'll submit following PR to address them.
Now let's merge this PR for coming demo session.

Issue #377, opened 2022-06-16
Feathr Documentation Table Render Issue with "Just the Doc"


Yuqing-cat:
https://linkedin.github.io/feathr/how-to-guides/feathr-configuration-and-env.html
![image](https://user-images.githubusercontent.com/26561033/174059258-4230d28c-e91b-4303-8054-542cdff65973.png)

It looks good in GitHub:
![image](https://user-images.githubusercontent.com/26561033/174061205-8aff3ceb-0028-4905-a371-afab7aaee93d.png)

Yuqing-cat:
Acceptance Criteria:
1. Fix the table render problem 
2. Update the doc link from "https://github.com/linkedin/feathr/blob/41e7496b38c43af6d7f8f1de842f657b27840f6d/docs/how-to-guides/feathr-configuration-and-env.md#a-note-on-using-azure-key-vault-to-store-credentials" to "https://linkedin.github.io/feathr/how-to-guides/feathr-configuration-and-env.html"

xiaoyongzhu:
confirm it's closed: 

<img width="986" alt="image" src="https://user-images.githubusercontent.com/1185090/178106437-f5791147-f9e5-4cf3-8497-d9fc854a9544.png">

Feel the root cause is the original table format isn't using a very standard way in markdown and justthedocs might not understand it.

PR #403, opened 2022-06-27
Documentation update & fix

xiaoyongzhu:
This PR fixes typos in the docs, update the formats for many of the tables (hopefully can fix https://github.com/linkedin/feathr/issues/377). This document fix doesn't include content changes.


blrchen (dismissed):

Yuqing-cat (commented):
Not quite understand the `when would you need a feature store` section. It may let users think feature store can only host one use case, which is not true. 
And for the face recognition sample, is it trying to refer to a `single data source` use case? Why can't it leverage Feathr to manage its feature generation?

blrchen (approved):

Yuqing-cat (approved):

PR #404, opened 2022-06-27
New Purview Registry adapt Sql data model

YihuiGuo:
Based on https://github.com/linkedin/feathr/pull/387
Add purview registry.

Keep sql registry code unchanged to avoid conflict.
Conform to Registy interfaces, provide same method and data format.

NOTICE that :
This PR does not totally eliminate the risk of concurrent conflict. 
In old version, we update all entities like 
`purviewclient.upload_entities([entity_batch])
`
In this PR, several changes has been applied to shrink the time window for a concurrent race condition to happen:
1. For entity uploading, only one entity was filled into the batch. (This means, register 1 entity for 1 upload)
2. Check for entity existence before uploading by qualifeidName+typeName.  If the matching entity exists, we will further check if these entities are identical, if not (an "update" operation), registry api will return HTTP409 Conflict. 

Limitations:
By applying changes above to purview registry, the time window for a race condition to happen shrinks, but not eliminated. 
A complete change needs discussion with Purview team about recommended way to secure MVCC.

### DIFF
--- a/registry/purview-registry/registry/purview_registry.py
+++ b/registry/purview-registry/registry/purview_registry.py
@@ -92,7 +92,6 @@ class PurviewRegistry(Registry):
         return [self.get_entity(x,recursive) for x in ids]
 
     def get_entity_id(self, id_or_name: Union[str, UUID]) -> UUID:
-        print(id_or_name)
         try:
             id = _to_uuid(id_or_name)
             return id
@@ -229,7 +228,7 @@ class PurviewRegistry(Registry):
         project_entity = self.get_entity(project_id)
         anchor_entity = self.get_entity(anchor_id)
         anchor_feature_entity = self.get_entity(anchor_feature_entity.guid)
-        source_entity = self.get_entity(anchor_entity.attributes.source['guid'])
+        source_entity = self.get_entity(anchor_entity.id)
 
         project_contains_feature_relation = self._generate_relation_pairs(
             project_entity, anchor_feature_entity, Label_Contains)
--- a/registry/purview-registry/registry/purview_registry.py
+++ b/registry/purview-registry/registry/purview_registry.py
@@ -1,12 +1,14 @@
 
+from http.client import CONFLICT, HTTPException
 import itertools
 import re
-from typing import Optional, Tuple, Union
+from typing import Any, Optional, Tuple, Union
+from urllib.error import HTTPError
 from uuid import UUID
 
 from azure.identity import DefaultAzureCredential
 from loguru import logger
-from numpy import allclose
+from numpy import allclose, typename
 from pyapacheatlas.auth.azcredential import AzCredentialWrapper
 from pyapacheatlas.core import (AtlasEntity, AtlasProcess,
                                 PurviewClient)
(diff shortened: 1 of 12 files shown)


registry/purview-registry/registry/purview_registry.py
```diff
@@ -0,0 +1,467 @@
...
+        """
+        Register the feathr types if we haven't done so. Note that this only needs to be called once per provisioning
+        a system. Basically this function registers all the feature type definition in a Atlas compatible system.
+        """
+
+        # Each feature is registered under a certain Feathr project. The project should what we refer to, however for backward compatibility, the type name would be `feathr_workspace`
+        type_feathr_project = EntityTypeDef(
+            name=str(EntityType.Project),
+            attributeDefs=[
+                # "anchor_features" and "derived_features" are removed, since we are moving to use process entity
```
xiaoyongzhu, line 336: Can we not remove those fields from the definition? We can just ignore those fields, as removing those attributes will need end users completely purge all the existing purview entities.

registry/purview-registry/registry/purview_registry.py
```diff
@@ -0,0 +1,467 @@
...
+            superTypes=["DataSet"],
+        )
+
+        type_feathr_derived_features = EntityTypeDef(
+            name=str(EntityType.DerivedFeature),
+            attributeDefs=[
+                AtlasAttributeDef(name="type", typeName="string",
+                                  cardinality=Cardinality.SINGLE),
+                # "input_anchor_features" and "input_derived_features" are deleted, use process entity instead
+                AtlasAttributeDef(name="key", typeName="array<map<string,string>>",
```
xiaoyongzhu, line 384: Can we not remove those fields from the definition? We can just ignore those fields, as removing those attributes will need end users completely purge all the existing purview entities.

registry/purview-registry/registry/purview_registry.py
```diff
@@ -0,0 +1,467 @@
...
+                AtlasAttributeDef(name="tags", typeName="map<string,string>",
+                                  cardinality=Cardinality.SINGLE),
+            ],
+            superTypes=["DataSet"],
+        )
+
+        type_feathr_anchors = EntityTypeDef(
+            name=str(EntityType.Anchor),
+            attributeDefs=[
+                # "source" will be removed, use process entity instead
```
xiaoyongzhu, line 397: Can we not remove those fields from the definition? We can just ignore those fields, as removing those attributes will need end users completely purge all the existing purview entities.

registry/purview-registry/registry/purview_registry.py
YihuiGuo: Ok, will put empty properties instead

registry/purview-registry/registry/purview_registry.py
YihuiGuo: Ok, will put empty properties instead

registry/purview-registry/registry/purview_registry.py
YihuiGuo: Ok, will put empty properties instead

blrchen (approved):

Yuqing-cat (approved):
Please make more e2e test

PR #409, opened 2022-06-28
Access Control Plugin for Registry APIs

Yuqing-cat:
This PR covers

- A access control plugin for all registry APIs (regardless of registry type)
- Access Control Management APIs and management UI
- A readme document

image


registry/api.py
```diff
@@ -0,0 +1,99 @@
...
+from fastapi import APIRouter, Depends
+from requests import request
+import requests
+from access_control.access import global_admin_access, get_user, project_read_access, project_write_access
+from access_control.models import User
+from access_control.db_rbac import DbRBAC
+
+router = APIRouter()
+rbac = DbRBAC()
+registry_url = "https://feathr-sql-registry.azurewebsites.net/api/v1"
```
blrchen, line 15: Please make this a server side configuration

registry/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.9
+
+COPY ./ /usr/src
+
+WORKDIR /usr/src
+RUN pip install -r requirements.txt
+
+# Start web server
+CMD [ "uvicorn","main:app","--host", "0.0.0.0", "--port", "80" ]
```
blrchen, line 9: I remember RBAC service and API service should share same docker image, @windoze is this correct?

registry/access_control/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+**Access Control Gateway** is an access control **Plugin** component of feature registry API. It can work with different type of backend registry. When user enables this component, registry requests will be validated in a gateway as below flow chart:
+
+```mermaid
+flowchart TD
+    A[Get Registry API Request] --> B{Is Id Token Valid?};
+    B -- No --> D[Return 401];
+    B -- Yes --> C{Have Permission?};
+    C -- No --> F[Return 403];
+    C -- Yes --> E[Call Registry API*];
```
blrchen, line 13: >Registry [](http://example.com/codeflow?start=24&length=8)

Call it downstreaming api might make more sense as new APIs like metrics, jobs will come soon

registry/api.py
Yuqing-cat: move to config.REGISTRY_URL

registry/access_control/README.md
Yuqing-cat: Updatd to "Downstream API"

registry/Dockerfile
Yuqing-cat: @windoze will help to package the docker image for ui, access control plugin, multi-option registry.

Yuqing-cat:
Some updates:
- Offer 2 options for @router.get("/features/{feature}":
     - if user pass `project` name, will validate access directly
     - if user do NOT pass `project` name, will validate access with the project name in feature response json
- Clearer configuration
    - Add ".env" sample file
    - update os.envrion[""] to os.environ.get() 
- Add Exception for requests without token
- The behavior of UI performs the same with test registry endpoint `https://feathr-sql-registry.azurewebsites.net` and `localhost` in all pages.


Will include in the following PR:
- Add a redirect page for 403 Exceptions to tell user what access is required.

registry/Dockerfile
blrchen: Then let's remove this file.

registry/access_control/db_rbac.py
```diff
@@ -0,0 +1,110 @@
...
+    def _get_userroles(self) -> list[UserRole]:
+        """query all the active user role records in SQL table
+        """
+        rows = self.conn.query(
+            fr"""select record_id, project_name, user_name, role_name, create_by, create_reason, create_time, delete_by, delete_reason, delete_time
+            from userroles
+            where delete_reason is null""")
+        ret = []
+        for row in rows:
+            r = UserRole(**row)
```
blrchen, line 30: >r = UserRole(**row) [](http://example.com/codeflow?start=12&length=19)

this line can be removed

registry/access_control/db_rbac.py
```diff
@@ -0,0 +1,110 @@
...
+class DbRBAC(RBAC):
+    def __init__(self):
+        if not os.environ.get("CONNECTION_STR"):
+            os.environ["CONNECTION_STR"] = config.CONNECTION_STR
+        self.conn = connect()
+        # cached user role records
+        self._refresh_cache()
+
+    def _refresh_cache(self):
+        #TODO: add user role cache refresh interval, e.g. daily
```
blrchen, line 17: > #TODO: add user role cache refresh interval, e.g. daily [](http://example.com/codeflow?start=7&length=56)

Do we really need this TODO?

registry/common/database.py
```diff
@@ -0,0 +1,153 @@
...
+                logging.warning("Database error, retrying...")
+                # Reconnect
+                self.make_connection()
+                retry += 1
+                if retry >= 3:
+                    # Stop retrying
+                    raise
+                pass
+
+    def update(self, sql: str, *args, **kwargs):
```
blrchen, line 89: It seems werid to make this method named update

registry/access_control/db_rbac.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def init_userrole(self, creator_name: str, project_name: str):
+        """initialize user role relationship when a new project is created
+        TODO: Add init user role to every new project call
+        """
+        create_by = "system"
+        create_reason = "creator of project, get admin by default."
+        query = fr"""insert into userroles (project_name, user_name, role_name, create_by, create_reason, create_time)
+            values ('{project_name}','{creator_name}','{RoleType.ADMIN}','{create_by}','{create_reason}', getutcdate())"""
+        return self.conn.update(query)
```
blrchen, line 110: >update [](http://example.com/codeflow?start=25&length=6)

Should trigger cache refresh here

registry/access_control/db_rbac.py
Yuqing-cat: This is TODO is to catch up the manual data insert into the `userroles` SQL table. 
We could do regular refresh as well as a manual refresh button on UI. What do you think?

registry/access_control/db_rbac.py
Yuqing-cat: Make sense.

registry/access_control/db_rbac.py
windoze: Cache doesn't work when RBAC service is in a cluster. And manually updating database is **not** a supported scenario.
We can skip cache and read/write database directly for now to make sure every operation is correct. Let's deal with performance later.

registry/common/database.py
Yuqing-cat: will modify function name in follow-up PR

registry/access_control/db_rbac.py
Yuqing-cat: Cache logic removed

registry/access_control/db_rbac.py
Yuqing-cat: Updated. Now every operation will read from database rather than refer to a cache

windoze (approved):

blrchen (approved):

PR #423, opened 2022-07-01
UI: Display feature version in node title if detected

blrchen:
This PR updates feature lineage graph page, to display feature version in node title if detected.


windoze (commented):
Not only in graph, but also in everywhere which needs to display feature/source/anchor names needs to display version to avoid confusion

Yuqing-cat (approved):

windoze (approved):

PR #426, opened 2022-07-02
Use SQL parameters instead of string concatenation

windoze:
Use SQL parameters instead of string concatenation to prevent SQL injection.


Yuqing-cat (approved):
Shall I follow this practice in access control? https://github.com/linkedin/feathr/pull/409

windoze:
> Shall I follow this practice in access control? #409

Yes, you should follow the practice as long as the parameters are passed from external untrusted source.

blrchen (approved):
:shipit:

PR #438, opened 2022-07-06
Update access control config name to unique ones

Yuqing-cat:
Background:
Configs like `AZURE_CLIENT_ID`, `API_BASE`, is commonly used in backend APIs. 
To package them in one docker image and enable different configurations for different components requires unique configs for each component.

Solution:
Add a common prefix "RBAC_" to all access control configs


windoze (dismissed):

Yuqing-cat:
- All access control related contents are moved under `access_control` folder.
- `x-registry-requestor` will be used to pass username to downstream APIs in header. 
- numpy dependency removed.

Tested in Local UI + SQL Registry. with all functions.

blrchen (approved):

windoze (approved):

PR #442, opened 2022-07-07
Use precondition to check mvcc conflict

YihuiGuo:
Per offline sync with Purview Team ,ETag (LastModifiedTS) can be used to perform MVCC. 
If the current Entity has ETag N on cloud side, then any updates with Etag<=N will be rejected. 

But not HTTP 409 Conflict will be returned, instead purview returns HTTP412 Precondition Failed. 

In our scenario, since we don't support entity update, every uploading should be considered as creation, so we could use a trick to manually set the payload ETag to 0, then if the entity exists on cloud side, Purview will return 412. Then we don't need to follow fetch-compare-upload pattern, and eliminate the small time window for a race condition to happen.

### DIFF
--- a/registry/purview-registry/registry/purview_registry.py
+++ b/registry/purview-registry/registry/purview_registry.py
@@ -497,15 +497,14 @@ class PurviewRegistry(Registry):
             force_update=True)
         logger.info("Feathr Feature Type System Initialized.")
 
-    def _upload_entity_batch(self, entity_batch):
+    def _upload_entity_batch(self, entity_batch:list[AtlasEntity]):
+        # we only support entity creation, update is not supported. 
+        # setting lastModifiedTS ==0 will ensure this, if another entity with ts>=1 exist
+        # upload funtion will fail with 412 Precondition fail.
         for entity in entity_batch:
-            logger.info(f"Creating {entity.qualifiedName} \t ({entity.typeName})")
-            existing_entity = self.purview_client.get_entity(qualifiedName=entity.qualifiedName, typeName=entity.typeName)
-            if existing_entity:
-            # perform attribute check, return id if same, conflict when different.
-                raise HTTPException(CONFLICT,"Entity with same qualified name and type exists")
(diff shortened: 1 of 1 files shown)


xiaoyongzhu (approved):

Yuqing-cat (approved):

registry/purview-registry/registry/purview_registry.py
```diff
@@ -497,15 +497,14 @@ def _register_feathr_feature_types(self):
             force_update=True)
         logger.info("Feathr Feature Type System Initialized.")
 
-    def _upload_entity_batch(self, entity_batch):
+    def _upload_entity_batch(self, entity_batch:list[AtlasEntity]):
+        # we only support entity creation, update is not supported. 
+        # setting lastModifiedTS ==0 will ensure this, if another entity with ts>=1 exist
+        # upload funtion will fail with 412 Precondition fail.
```
garyzjq, line 503: is there any code to handle 412 exception?

blrchen, line 503: Since 412 is user error, feathr client should explicitly tell user error details on the operation fails. Show message like The specified feature <> already exists. Probably sql registry already has an error code for this.

YihuiGuo: Ok, will check sql error handling and try raise some right error code. 
Currently on API caller side it is just 500

YihuiGuo: This will be running in a standalone API project, on the api caller side, information will be 500. 
Will refine this to show 412 to the API caller, thanks for the advice! :P

jainr (approved):

registry/purview-registry/registry/purview_registry.py
YihuiGuo: The plan to fix this is merge this first, then Chen will add unified error handling to expose 412 to the customer.

PR #443, opened 2022-07-07
Enhancing the deployment experience

jainr:
1. Currently the user needs to pass the Principal Id as a pre-req, we now create a managed identity within the template.
2. We also removed the bacpac URL paramter from UI, since that is something user should not need to worry about

Successful Deployment 
image


xiaoyongzhu (dismissed):

blrchen:
Hello @jainr , I tried with this PR and the web app failed to start, could you please make following changes to make it align latest code

1. Change name of app settings to following two, now UI and backend use different client id. These two are generated on the flying after ui app registration is completed, user does not need to provide
  - REACT_APP_AZURE_CLIENT_ID
  - REACT_APP_AZURE_TENANT_ID
2. Change docker image name to feathrfeaturestore/feathr-registry. Now latest docker image supports selected registry scenario.

jainr:
Thanks Blair, I addressed your comments.

edwincheung (approved):

Yuqing-cat (approved):

Issue #46, opened 2022-03-10
Add developer related docs


xiaoyongzhu:
We should have developer focused docs for:
- CI pipeline
- integration with Azure resources
- Feathr internals

PR #450, opened 2022-07-11
update config for better deployment experience

Yuqing-cat:
This PR contains some minor updates to improve deployment experience:
- Update `API_BASE` config to follow the same pattern of registry APIs
- Set default value of `RBAC_API_AUDIENCE` to `RBAC_API_CLIENT_ID`
- Update `403` Error log for better debug experience


registry/access_control/rbac/auth.py
```diff
@@ -115,14 +115,14 @@ def _decode_token(self, token: str) -> Mapping:
                                 'RS256'], audience=config.RBAC_API_AUDIENCE)
             return decode
         except ExpiredSignatureError as e:
-            logging.debug(f'The token signature has expired: {e}')
+            logging.error(f'The token signature has expired: {e}')
```
blrchen, line 118: ExpiredSignatureError is client error, should use logging.info instead on server side.

registry/access_control/rbac/auth.py
```diff
@@ -115,14 +115,14 @@ def _decode_token(self, token: str) -> Mapping:
...
             raise InvalidAuthorization('The token signature has expired')
         except PyJWKError as e:
-            logging.debug(f'Invalid token: {e}')
+            logging.error(f'Invalid token: {e}')
             raise InvalidAuthorization('The token is invalid')
         except Exception as e:
-            logging.debug(f'Unexpected error: {e}')
-            raise InvalidAuthorization('Unable to decode token, error: {e}')
+            logging.error(f'Unexpected error: {e}')
+            raise InvalidAuthorization(f'Unable to decode token, error: {e}')
```
blrchen, line 125: >raise InvalidAuthorization(f'Unable to decode token, error: {e}') [](http://example.com/codeflow?start=12&length=65)

Unable to decode token indicates server error, should return 500 internal server error instead.

Yuqing-cat:
Will fix in follow up PRs:
- raise Exception when environment variables are not set. https://github.com/linkedin/feathr/issues/46

blrchen (approved):

windoze (approved):

Issue #422, opened 2022-06-30
Empty values in Product Recommendation Demo


t-curiekim:
All the values for the feature "feature_user_purchasing_power" is empty for the demo notebook **product_recommendation_demo.ipynb**. The demo notebook was run as it is without any changes made to it.

<img width="265" alt="Screen Shot 2022-06-30 at 1 20 00 PM" src="https://user-images.githubusercontent.com/105448774/176771292-45d00375-8430-459d-9ff9-7d88de7a04da.png">

The empty value problem does not occur when treating the boolean feature as a string like below, but the if expression always resolves to false

> transform="feature_user_gift_card_balance + if_else(feature_user_has_valid_credit_card == 'True', 100, 0)"

PR #458, opened 2022-07-12
[feathr] Fix if_else bug in notebook

hangfei:
FIx for https://github.com/linkedin/feathr/issues/422


xiaoyongzhu (approved):

blrchen (approved):
Approved with an open question for discussion.

Currently this error cannot be caught in both build time and run time. The generated spark sql is good, the if_else UDF can accept a non bool type as condition with no error. It seems the only way to discover this error is manual verify published data. 

How can this experience be improved?

PR #467, opened 2022-07-14
Minor fixes and feedback from Bugbash

jainr:
- This PR contains minor fixes based on feedback from deployment bugbash.
- Removing the deployment instructions from README.md to avoid duplicates and have single source of truth. https://github.com/linkedin/feathr/blob/main/docs/how-to-guides/azure-deployment.md
- Updated autoscaling min and max node for Synapse, since min node count can't be smaller than 3
- Added support for RBAC schema import in the template.
- Separated documentation into CLI and ARM as it was confusing within single doc.
- More updates to documentation based on bug bash feedback.
- Removed role assignment for Blob Contributor from ARM template as it is not required and have to be added in the notebook. Since deployment identity and notebook user identity will be different.
- Added workaround for SQL import.
=======

### There will be an upcoming PR with more fixes which has been identified in testing.

- Adding instructions in doc to add roles for collection Admin and Data Curator for User Assigned Managed identity in Azure Purview.
- Add User Assigned managed identity to webapp so it can communicate with Purview.
- Bug - Import can only work on empty db, so two imports that current PR has, one job would always fail. We are merging two bacpacs into one so only one import is done.


edwincheung (approved):

docs/README.md
```diff
@@ -31,34 +31,8 @@ Feathr automatically computes your feature values and joins them to your trainin
 
 ## Running Feathr on Azure with few Simple Steps
 
-1. To enable authentication on the Feathr UI (which gets created as part of the deployment script) we need to create an Azure Active Directory (AAD) application. Currently it is not possible to create one through ARM template but you can easily create one by running the following CLI commands in the [Cloud Shell](https://shell.azure.com/bash)
+Follow the [feathr deployment guide ](./how-to-guides/azure-deployment.md) to run feathr on Azure.
```
Yuqing-cat, line 34: A suggested way to link to documents are pointing to document site: 
`[feathr deployment guide](https://linkedin.github.io/feathr/how-to-guides/azure-deployment.html)`
[feathr deployment guide](https://linkedin.github.io/feathr/how-to-guides/azure-deployment.html)

blrchen, line 34: >feathr [](http://example.com/codeflow?start=82&length=6)

feathr => Feathr to align the rest of doc

docs/how-to-guides/azure_resource_provision.json
```diff
@@ -380,9 +391,16 @@
...
               "value": "[concat('Server=tcp:',reference(variables('sqlServerName')).fullyQualifiedDomainName,',1433;Initial Catalog=',variables('sqlDatabaseName'),';Persist Security Info=False;User ID=',reference(variables('sqlserverName')).administratorLogin,';Password=',parameters('sqlAdminPassword'),';MultipleActiveResultSets=False;Encrypt=True;TrustServerCertificate=False;Connection Timeout=30;')]"
+            },
+            {
+              "name": "ENABLE_RBAC",
+              "value": "[if(equals(parameters('enableRBAC'),'Yes'),json('true'),json('null'))]"
+            },
+            {
+              "name": "PURVIEW_NAME",
+              "value": "[if(equals(parameters('registryBackend'),'Azure-Purview'),variables('purviewName'),json('null'))]"
             }
```
blrchen, line 404: When Purview is selected as registry provider, AZURE_CLIENT_ID, AZURE_TENANT_ID, AZURE_CLIENT_SECRET are required, these is the service principal registry server uses to talk to purview.

docs/README.md
jainr: Interesting, I didn't know that, is there a preference of one over the other?

xiaoyongzhu: Since all the docs now are in the docs/ folder, the doc builder will automatically convert those relative paths to absolute ones (so the ones that end users see will be `https://linkedin.github.io/feathr/how-to-guides/azure-deployment.html`

windoze (approved):

PR #468, opened 2022-07-14
support aad application tokens

Yuqing-cat:
This PR extend the token decode functions to 
- Support `AAD Application` by using `appid` as username
- Support other Oauth tokens by
    - changing required id from `oid` to `sub`
    - adding 2 user types with `email` / `sub` as username 
- Update readme

### DIFF
--- a/registry/access_control/README.md
+++ b/registry/access_control/README.md
@@ -40,13 +40,15 @@ Admin roles can add or delete roles in management UI page or through management 
 
 ### Environment Settings
 
+`ENABLE_RBAC` needs to be set to deploy a registry backend with access control plugin.
+
 | Variable| Description|
 |---|---|
 | RBAC_CONNECTION_STR| Connection String of the SQL database that host access control tables, required.|
-| RBAC_API_BASE| Aligned API base, needs to start with `/`|
+| RBAC_API_BASE| Aligned API base|
 | RBAC_REGISTRY_URL| The downstream Registry API Endpoint|
-| RBAC_AAD_INSTANCE | Set to "https://login.microsoftonline.com" by default |
-| RBAC_AAD_TENANT_ID| Used get auth url together with `RBAC_AAD_INSTANCE`, set to `common` by default|
+| RBAC_AAD_INSTANCE | Instance like "https://login.microsoftonline.com" |
+| RBAC_AAD_TENANT_ID| Used get auth url together with `RBAC_AAD_INSTANCE`|
 | RBAC_API_AUDIENCE| Used as audience to decode jwt tokens|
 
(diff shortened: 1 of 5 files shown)


registry/access_control/rbac/auth.py
```diff
@@ -50,14 +48,37 @@ async def __call__(self, request: Request) -> User:
     @staticmethod
     def _get_user_from_token(decoded_token: Mapping) -> User:
         try:
-            user_id = decoded_token['oid']
+            user_id = decoded_token['sub']
         except Exception as e:
-            raise InvalidAuthorization(detail=f'Unable to extract user details from token, {e}')
+            raise InvalidAuthorization(detail=f'Unable to extract subject as unique id from token, {e}')
+
+        print(decoded_token)
```
blrchen, line 55: >decoded_token [](http://example.com/codeflow?start=14&length=13)

token is sensitive, should not print/log

registry/access_control/README.md
```diff
@@ -40,13 +40,15 @@ Admin roles can add or delete roles in management UI page or through management
 
 ### Environment Settings
 
+`ENABLE_RBAC` needs to be set to deploy a registry backend with access control plugin.
```
xiaoyongzhu, line 43: Can we also add those envs in the centralized environment variable doc, so it's easier for users to track/

Yuqing-cat: This setting is included in the registry deployment that @jainr is working on. I think he already created a doc for it.  But we would merge those environment variables into one place in a following PR.

blrchen: The environment variable set for deployment is different than the set for Feathr client users. For example, people who owns deployment should if RBAC should be enabled or not whereas Feathr client user might not need to be aware of this env.

Since the two environment variable sets are for different purpose, I suggest we document these in different docs.

blrchen: I think it's better to remove ENABLE_RBAC description in this README, as it's only consumed by deployment time, not by RBAC application runtime. Put it in RBAC dev doc will introduce confusion.

xiaoyongzhu: > The environment variable set for deployment is different than the set for Feathr client users. For example, people who owns deployment should if RBAC should be enabled or not whereas Feathr client user might not need to be aware of this env.
> 
> Since the two environment variable sets are for different purpose, I suggest we document these in different docs.

I see, thanks!

xiaoyongzhu (approved):

xiaoyongzhu:
LGTM but would be better to rely on @blrchen 's opinions here

blrchen (approved):

PR #475, opened 2022-07-15
Generic Input/Output of DataFrames

windoze:
This PR includes:
1. Extended former `InputLocation` classes to support both read and write functions, also renamed it to `DataLocation` to reflect this change.
2. Added a `GenericLocation` which supports all Spark confs, modes, and options, so it can be used to operate virtually any connectors supported by Spark
3. In `GenericLocation`, I added a format-specific patching mechanism to workaround quirks come with different connectors, e.g. CosmosDb requires rows to have an `id` column with unique values.
4. Update `FeathrGenJob` and `FeathrJoinJob`, enabling them to use a JSON-encoded string of `DataLocation` instead of plain `path` as the input and output target.

Theoretically, Feathr core can support all Spark connectors with this patch, but we still run a series of compatibility tests to confirm the final list.

NOTE: This PR only involves Feathr core, the corresponding Feathr Client changes will be in upcoming PRs.


build.sbt
```diff
@@ -44,7 +44,9 @@ val localAndCloudCommonDependencies = Seq(
     "net.snowflake" % "spark-snowflake_2.12" % "2.10.0-spark_3.2",
     "org.apache.commons" % "commons-lang3" % "3.12.0",
     "org.xerial" % "sqlite-jdbc" % "3.36.0.3",
-    "com.github.changvvb" %% "jackson-module-caseclass" % "1.1.1"
+    "com.github.changvvb" %% "jackson-module-caseclass" % "1.1.1",
+    "com.azure.cosmos.spark" % "azure-cosmos-spark_3-1_2-12" % "4.11.1",
+    "org.eclipse.jetty" % "jetty-util" % "9.3.24.v20180605"
```
xiaoyongzhu, line 49: Do we need org.eclipse.jetty here?

xiaoyongzhu:
This PR looks good to me and is a good way of extending to other sources in the future. Thanks @windoze for the work!

build.sbt
```diff
@@ -44,7 +44,9 @@ val localAndCloudCommonDependencies = Seq(
     "net.snowflake" % "spark-snowflake_2.12" % "2.10.0-spark_3.2",
     "org.apache.commons" % "commons-lang3" % "3.12.0",
     "org.xerial" % "sqlite-jdbc" % "3.36.0.3",
-    "com.github.changvvb" %% "jackson-module-caseclass" % "1.1.1"
+    "com.github.changvvb" %% "jackson-module-caseclass" % "1.1.1",
+    "com.azure.cosmos.spark" % "azure-cosmos-spark_3-1_2-12" % "4.11.1",
```
xiaoyongzhu, line 48: Can we use `com.azure.cosmos.spark:azure-cosmos-spark_3-2_2-12:4.11.2` if possible, assuming this cosmosdb 3.2 package can work well with Spark 3.1?

windoze: Unfortunately no. Spark 3.1 and 3.2 are not that compatible.

I'm thinking of an improvement to add the jar dependency from the client side, so we can remove this dep from the core.
But then the client needs the knowledge of the cluster runtime version before submitting the job, I'm still trying to figure out a clear way to do that.
Or we can create 2 different Feathr JARs, one for Spark 3.1 and the other for 3.2, that's also the standard way that other Spark-related projects followed.

xiaoyongzhu: interesting... in that case, maybe we can use Maven for those jars?

i.e. don't package them in the fat Feathr jar, but when submitting the job, always use Maven to get  `com.azure.cosmos.spark`, and as you said, in the client side users can specify the spark version?

windoze: Both DataBricks and Synapse support specifying Spark version, but in very different ways. In DataBricks user can set the DataBricks Runtime version in the job submission request if `new_cluster` is used, but in Synapse the Spark version is set when creating the cluster, and cannot be changed later.
If we want the client to set JAR deps, it may need a solid way to detect all Spark cluster info and compose the dep list from it, but I'm afraid this procedure could be pretty error-prone.

src/main/scala/com/linkedin/feathr/offline/config/location/GenericLocation.scala
```diff
@@ -0,0 +1,161 @@
...
+            case None => {
+              // If there is no key column, we use a auto-generated monotonic id.
+              // but in this case the result could be duplicated if you run job for multiple times
+              // This function is for offline-storage usage, ideally user should create a new container for every run
+              df.withColumn("id", (monotonically_increasing_id().cast("string")))
+            }
+          }
+        } else {
+          // We already have an `id` column
+          // TODO: Should we do anything here?
```
jaymo001, line 142: Clean up

src/main/scala/com/linkedin/feathr/offline/config/location/GenericLocation.scala
```diff
@@ -0,0 +1,161 @@
...
+              // Generate key column from header info, which is required by CosmosDb
+              val (keyCol, keyedDf) = DataFrameKeyCombiner().combine(df, FeatureGenUtils.getKeyColumnsFromHeader(h))
+              // Rename key column to `id`
+              keyedDf.withColumnRenamed(keyCol, "id")
+            }
+            case None => {
+              // If there is no key column, we use a auto-generated monotonic id.
+              // but in this case the result could be duplicated if you run job for multiple times
+              // This function is for offline-storage usage, ideally user should create a new container for every run
+              df.withColumn("id", (monotonically_increasing_id().cast("string")))
```
jaymo001, line 137: There's a potential issue with monotonically_increasing_id() in Spark. i.e. when some partition of the RDD is recalculated under the hood, it may produce different values from the first run for the same partition resulting in duplicates across different partitions. A cache might be necessary in such cases.

src/main/scala/com/linkedin/feathr/offline/config/location/GenericLocation.scala
```diff
@@ -0,0 +1,161 @@
...
+      .load()
+  }
+
+  def writeDf(ss: SparkSession, df: DataFrame, header: Option[Header], location: GenericLocation) = {
+    location.conf.foreach(e => {
+      ss.conf.set(e._1, e._2)
+    })
+
+    location.format.toLowerCase() match {
+      case "cosmos.oltp" =>
```
jaymo001, line 112: Separate this out to its own class/interface is preferred.

src/main/scala/com/linkedin/feathr/offline/config/location/GenericLocation.scala
```diff
@@ -0,0 +1,161 @@
...
+      options += (key.replace("__", ".") -> LocationUtils.envSubstitute(v))
+    }
+  }
+}
+
+/**
+ * Some Spark connectors need extra actions before read or write, namely CosmosDb and ElasticSearch
+ * Need to run specific fixes base on `format`
+ */
+object GenericLocationFixes {
```
jaymo001, line 96: Can we find a better name than "Fixes"?

src/main/scala/com/linkedin/feathr/offline/config/location/DataLocation.scala
```diff
@@ -86,3 +104,50 @@ object LocationUtils {
...
+      case _ @ (_: ConfigException | _: JacksonException) => SimplePath(cfg)
+    }
+  }
+
+  def apply(cfg: Config): DataLocation = {
+    apply(cfg.root().keySet().asScala.map(key ⇒ key → cfg.getString(key)).toMap)
+  }
+
+  def apply(cfg: Any): DataLocation = {
+    val jackson = (new ObjectMapper(new HoconFactory) with CaseClassObjectMapper)
```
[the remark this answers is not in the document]
windoze: HOCON is a superset of JSON, every valid JSON is valid HOCON.
The current pattern is always valid if the config is generated by Feathr client, it works unless we need to support hand-written config files.

src/main/scala/com/linkedin/feathr/offline/config/location/GenericLocation.scala
windoze: Do we have any conclusion in this case? It's better to leave this `TODO` if we don't, for tracking purpose.

src/main/scala/com/linkedin/feathr/offline/config/location/GenericLocation.scala
windoze: monotonically_increasing_id is constructed by partition id and row id in partition, it keeps the same even if the RDD has been re-processed.
Besides, we're considering if we should allow users to use CosmosDb as the offline store, it may not make much sense after all.

src/main/scala/com/linkedin/feathr/offline/config/location/GenericLocation.scala
windoze: These codes are just ad-hoc patches for different drivers and/or formats, maybe there isn't a better name unless we separate them into different subclasses, which seems too verbose.

src/main/scala/com/linkedin/feathr/offline/config/location/GenericLocation.scala
windoze: The separation involves another class hierarchy which may not be solely determined by the `format` field on deserialization.
The plan is to support AeroSpike and maybe ElasticSearch in the next stage, I'll try to clean them up at that time.

Yuqing-cat (approved):

blrchen (approved):

PR #479, opened 2022-07-18
Update Dockerfile to point to current nginx.conf file

donegjookim:
I discovered that the Docker build fails from following the Azure quickstart guide. The error stems from incorrect nginx.conf path:
image

I verified that the docker build works correctly after the fix.


blrchen:
Hello @donegjookim, thanks for fixing this. This error was not caught earlier because docker image publishing was only tested on windows.

I spent some time test this image on linux and observed one more fix is required, otherwise when you perform `docker run` with build image, you will experience `error /bin/sh: 1: ./start.sh: Permission denied`

Would you mind to include this as part of PR? Its one line change, adding following before line 29
`RUN ["chmod", "+x", "./start.sh"]`

PR #483, opened 2022-07-19
Add prettier and integrate it with eslint

donegjookim:
The main goal of this PR is to integrate prettier with eslint, precommit hook, and git hub action. Besides that, I

- Added eslint check in github action for consistency
- Added React custom hooks linting rule to ensure best practices
- Added .git-blame-ignore-revs to ensure that git blame history does not get sabotaged


.prettierrc
```diff
@@ -0,0 +1 @@
+{}
```
blrchen, line 1: Please move prettierrc, .eslintignore files to under ui directory, as these files are scoped to ui code only.

Is .git-blame-ignore-revs scoped to all codes in repo? If no probably this file should get moved too.

.github/workflows/pull_request_push_test.yml
```diff
@@ -32,6 +32,28 @@ jobs:
           distribution: "temurin"
       - name: Run tests
         run: sbt clean && sbt test
+
+  eslint:
+    name: eslint
```
blrchen, line 37: >name: eslint [](http://example.com/codeflow?start=4&length=12)

Please add trigger condition so eslint and prettier formatter only fires when PR includes change for ui files.

ui/README.md
```diff
@@ -47,6 +52,7 @@ REACT_APP_AZURE_TENANT_ID=<REPLACE_WITH_YOUR_TENANT_ID>
 ### Lint
```
blrchen, line 52: Please add some contents under Lint section in README.md, so people can qucikly know what lints/code formatters have been configured. Doc should indicate where the rule are configured as well.

.prettierrc
donegjookim: Thanks for catching - I totally missed moving prettierrc. It's been a while since I worked with a mono repo.

donegjookim: As it is related to git, .git-blame-ignore-revs is per git repo basis.
It makes sense to be at the root directory as we're only using one git repository

ui/README.md
donegjookim: I added comment about Prettier and mentioned that we're using the default rules.

.github/workflows/pull_request_push_test.yml
donegjookim: I created another file to address this issue - it'll better help to separate API checks and UI checks in the workflow for readability and maintainability

blrchen (approved):

Yuqing-cat (approved):

PR #500, opened 2022-07-25
Anchor feature node and source node should not have nagivate icon

enya0405:
- remove navigation icon and search function for source nodes and anchor nodes


enya0405:
Redundant PR

PR #501, opened 2022-07-25
remove extra navigation icon and add control for search button

enya0405:
- remove navigation icon for 'Source' nodes and 'Anchor' nodes
- disable 'Search' button when no project set


ui/src/components/graph/graph.tsx
```diff
@@ -38,6 +38,10 @@ const Graph: React.FC<Props> = ({ data, nodeId }) => {
     setElements(layoutedElements);
   }, [layoutedElements, data, nodeId]);
 
+  const isFeature = (featureType: string) => {
+    return featureType === 'feathr_anchor_feature_v1' || featureType === 'feathr_derived_feature_v1'
+  }
```
blrchen, line 43: There is already one isFeature in ui code, please try to reuse that. 

Can you pls also make 'feathr_anchor_feature_v1' an enum?

enya0405: - moved 'isFeature' into utils under graph/ so that other files can access it more easily
- make all nodes types as enums in utils

ui/src/components/graph/utils.ts
```diff
@@ -182,7 +182,22 @@ const generateEdge = ({ obj, from, to }: GenerateEdgeProps): Edge => {
   };
 };
 
-export { generateEdge, generateNode, getLayoutedElements, getElements };
+const enum NodeTypes {
```
blrchen, line 185: >NodeTypes [](http://example.com/codeflow?start=11&length=9)

It's better name it FeatureType so non graph code can uses it too. And this should be moved to utils under ui root directory for same purpose

enya0405: Modified as required.

blrchen (approved):

Yuqing-cat (approved):

PR #504, opened 2022-07-25
Do not trigger python & scale tests for pure ui changes

blrchen:
This PR updates CI/CD workflow to exclude ui code changes from python & scala tests, as UI code has a separated workflow ui-check.yml


jainr (approved):
Makes sense to me.

Yuqing-cat (approved):

PR #505, opened 2022-07-25
Increase azure Databricks job timeout in Azure integration tests

blrchen:
Currently many ci job fails due to Azure Databricks job takes a long time to complete. The Feathr client in E2E test has 15 minutes timeout so if ADB job takes longer than 15 minutes to complete, test will fail.

This PR doubles the Feathr client job timeout, to reduce the chance contributors seeing CI fail with infra issues.

@jainr could you pls co-investigate why ADB jobs take long to complete in CI env when you make improvements on CI/CD? For my own ADB cluster, those jobs normally take 6-7 minutes to complete.


jainr (dismissed):

Yuqing-cat (dismissed):
Will it be better to extract a default constant or a config for `default_spark_job_timeout_sec` rather than making change in every test script?

feathr_project/test/test_azure_spark_maven_e2e.py
```diff
@@ -42,7 +42,7 @@ def test_feathr_online_store_agg_features():
     client.materialize_features(settings)
     # just assume the job is successful without validating the actual result in Redis. Might need to consolidate
     # this part with the test_feathr_online_store test case
-    client.wait_job_to_finish(timeout_sec=900)
+    client.wait_job_to_finish(timeout_sec=1800)
```
xiaoyongzhu, line 45: maybe we should use `Constants.SPARK_JOB_TIMEOUT_SECONDS`

feathr_project/test/test_input_output_sources.py
```diff
@@ -50,7 +50,7 @@ def test_feathr_get_offline_features_with_parquet():
                                 )
 
     # assuming the job can successfully run; otherwise it will throw exception
-    client.wait_job_to_finish(timeout_sec=900)
+    client.wait_job_to_finish(timeout_sec=1800)
```
xiaoyongzhu, line 53: maybe we should use `Constants.SPARK_JOB_TIMEOUT_SECONDS`

feathr_project/test/test_input_output_sources.py
```diff
@@ -95,7 +95,7 @@ def test_feathr_get_offline_features_with_delta_lake():
...
```
xiaoyongzhu, line 98: maybe we should use `Constants.SPARK_JOB_TIMEOUT_SECONDS`

feathr_project/test/test_sql_source.py
```diff
@@ -130,7 +130,7 @@ def test_feathr_get_offline_features():
                                 output_path=output_path)
 
     # assuming the job can successfully run; otherwise it will throw exception
-    client.wait_job_to_finish(timeout_sec=900)
+    client.wait_job_to_finish(timeout_sec=1800)
```
xiaoyongzhu, line 133: maybe we should use `Constants.SPARK_JOB_TIMEOUT_SECONDS`

xiaoyongzhu:
This LGTM, thanks for taking time to fix those issues!

jainr (approved):

xiaoyongzhu (approved):

PR #511, opened 2022-07-27
Change feature registry endpoint and required env variables

t-curiekim:
Environment variables AZURE_CLIENT_ID, AZURE_TENANT_ID, and AZURE_CLIENT_SECRET are no longer needed, and the feature registry configuration should be updated to talk to the registry api endpoint verses talking directly to purview


Yuqing-cat (approved):

jainr (approved):

jainr:
What will be the configuration be if the user selected SQL registry?

blrchen:
> What will be the configuration be if the user selected SQL registry?

The endpoint url just use the web app created by deployment. The arm template determines the web app settings based on user's selection. Web app relays on the settings to decide which register to run ( sql vs purview), rbac on/off.

jainr:
> > What will be the configuration be if the user selected SQL registry?
> 
> The endpoint url just use the web app created by deployment. The arm template determines the web app settings based on user's selection. Web app relays on the settings to decide which register to run ( sql vs purview), rbac on/off.

Got it, I was confused. So Notebook points to registry URL --> Registry talks to Purview or SQL based on the App Settings.

PR #512, opened 2022-07-27
update token fetch for the registry and synapse

t-curiekim:
This PR contains changes to the following:

- DefaultAzureCredentials.get_token method returns an AccessToken object. the ".token" needs to be added to grab the string value of the token in the AccessToken object. 
- The ".default" url throws authentication error in the Feathr registry client when running the notebook, however, replacing the url with "https://dev.azuresynapse.net/.default" resolves the issue. More investigation is needed as to why.

Screen Shot 2022-07-22 at 2 14 41 PM

### DIFF
--- a/feathr_project/feathr/spark_provider/_synapse_submission.py
+++ b/feathr_project/feathr/spark_provider/_synapse_submission.py
@@ -326,7 +326,7 @@ class _SynapseJobRunner(object):
         # @see: https://docs.microsoft.com/en-us/azure/synapse-analytics/spark/connect-monitor-azure-synapse-spark-application-level-metrics
         app_id = self.get_spark_batch_job(job_id).app_id
         url = "%s/sparkhistory/api/v1/sparkpools/%s/livyid/%s/applications/%s/driverlog/stdout/?isDownload=true" % (self._synapse_dev_url, self._spark_pool_name, job_id, app_id)
-        token = self._credential.get_token("https://dev.azuresynapse.net/.default")
+        token = self._credential.get_token("https://dev.azuresynapse.net/.default").token
         req = urllib.request.Request(url=url, headers={"authorization": "Bearer %s" % token})
         resp = urllib.request.urlopen(req)
         return resp.read()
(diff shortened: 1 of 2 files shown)


xiaoyongzhu:
This is interesting as we are contacting purview but need to grab token from Azure Synapse?

feathr_project/feathr/spark_provider/_synapse_submission.py
```diff
@@ -326,7 +326,7 @@ def get_driver_log(self, job_id) -> str:
         # @see: https://docs.microsoft.com/en-us/azure/synapse-analytics/spark/connect-monitor-azure-synapse-spark-application-level-metrics
         app_id = self.get_spark_batch_job(job_id).app_id
         url = "%s/sparkhistory/api/v1/sparkpools/%s/livyid/%s/applications/%s/driverlog/stdout/?isDownload=true" % (self._synapse_dev_url, self._spark_pool_name, job_id, app_id)
-        token = self._credential.get_token("https://dev.azuresynapse.net/.default")
+        token = self._credential.get_token("https://dev.azuresynapse.net/.default").token
```
blrchen, line 329: >get_token("https://dev.azuresynapse.net/.default") [](http://example.com/codeflow?start=33&length=50)

Synapse token is for protecting Synapse resources only. Please change the scope to [https://management.azure.com/.default](https://management.azure.com/.default) which is the URI for ARM(azure resource management) which is more appreciable here

Yuqing-cat (commented):
Echo with Blair's comment, a related PR for rbac registry is https://github.com/linkedin/feathr/pull/514
`audience=["https://management.azure.com",config.RBAC_API_AUDIENCE]`

feathr_project/feathr/spark_provider/_synapse_submission.py
jainr: @t-curiekim tried with grap.azure.com url earlier and that didn't work as that as the same net effect as just called .default. Agree, let's try the management.azure.com URL and that might work.

t-curiekim: Have used the management azure url and seemed to work before. I will try testing a few more times with the management url to confirm.

t-curiekim: Thank you for testing! I have tried running again now, and works on my side as well 👍

jainr (approved):
Thanks Blair for pointing out the management URL, thanks Curie for the fix.

blrchen (approved):

PR #516, opened 2022-07-27
demo of how to enable coercion between different kinds of FeatureValue classes in MVEL expressions

d4ve:
This is a low-level demo of how we can set up MVEL to allow coercions of different kinds of FeatureValue classes. The motivation for this is to build a migration aid from previous versions of Feathr (from before open sourcing): we can use this capability in MVEL to build a plugin to allow running legacy feature definitions that contain MVEL expressions based on old versions of FeatureValue APIs.


xiaoyongzhu:
Close this PR as it is stale for a while. Let me know if anyone wants to reopen it.

PR #520, opened 2022-07-28
Rename Dockerfile to FeathrRegistry.Dockerfile

xiaoyongzhu:
Some online services such as mybinder will take the Dockerfile in the root directory and build it. This doesn't apply to Feathr as this docker file is used to build UI and the REST APIs. Rename it to avoid wrong assumptions.


blrchen:
Hello @xiaoyongzhu , existing docker build command `docker build -t feathrfeaturestore/feathr-registry` will break after this naming. I am ok with the docker file name change, I will submit docs update in this PR as well.

blrchen (approved):

Yuqing-cat (approved):

PR #531, opened 2022-07-29
DataPathHandler Bug Fix

blee1234:
Fixing a bug where external data paths don't accidentally trigger the HDFS path when executing local path logic.

Namely with LocalPathChecker, it needs to have logic to identify when a path isn't HDFS and thus will not invoke HdfsUtils.exists, which currently throws an exception when non HDFS local paths get passed in.

### DIFF
--- a/src/main/scala/com/linkedin/feathr/offline/source/pathutil/LocalPathChecker.scala
+++ b/src/main/scala/com/linkedin/feathr/offline/source/pathutil/LocalPathChecker.scala
@@ -1,13 +1,14 @@
 package com.linkedin.feathr.offline.source.pathutil
 
 import com.linkedin.feathr.offline.util.{HdfsUtils, LocalFeatureJoinUtils, SourceUtils}
+import com.linkedin.feathr.offline.source.dataloader.DataLoaderHandler
 import org.apache.hadoop.conf.Configuration
 
 /**
  * path checker for local test files.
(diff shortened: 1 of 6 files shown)


src/main/scala/com/linkedin/feathr/offline/source/pathutil/LocalPathChecker.scala
```diff
@@ -20,13 +21,33 @@ private[offline] class LocalPathChecker(hadoopConf: Configuration) extends PathC
     LocalFeatureJoinUtils.getMockPathIfExist(path, hadoopConf, None).isDefined
   }
 
+  /**
+   * check whether the input path is an Hdfs path. Neesd to have separate function, as there are class conflicts with Breaks.getClass
+   * @param path input path.
+   * @return true if the path is an Hdfs path.
+   */
+  def isHdfs(path: String): Boolean = {
```
jaymo001, line 29: From the implementation, it is not related to HDFS, but whether it has handler or not. Change to a better name?

blee1234: sg, changed it isExternalDataSource.

jaymo001 (approved):

xiaoyongzhu (approved):

PR #545, opened 2022-08-02
Support 'enabled' configurations for offline stores

enya0405:
For offline stores, S3/ADLS/WASB/JDBC/Snowflake, if any of its 'enabled' config variables are set to 'false' or undefined their other related environment variables will not be visited.


xiaoyongzhu:
Quick comment - instead just return empty string, we should also eliminate the parameter names below:

```
                '--s3-config', self._get_s3_config_str(),
                '--adls-config', self._get_adls_config_str(),
                '--blob-config', self._get_blob_config_str(),
                '--sql-config', self._get_sql_config_str(),
                '--snowflake-config', self._get_snowflake_config_str()
```

i.e. if s3 is not configured, it doesn't make sense to have `--s3-config` in the param list

enya0405:
> Quick comment - instead just return empty string, we should also eliminate the parameter names below:
> 
> ```
>                 '--s3-config', self._get_s3_config_str(),
>                 '--adls-config', self._get_adls_config_str(),
>                 '--blob-config', self._get_blob_config_str(),
>                 '--sql-config', self._get_sql_config_str(),
>                 '--snowflake-config', self._get_snowflake_config_str()
> ```
> 
> i.e. if s3 is not configured, it doesn't make sense to have `--s3-config` in the param list

Modified the solution in new commit.

jainr (dismissed):
Thanks for the update, changes look good to me.

xiaoyongzhu (approved):

blrchen (approved):

PR #555, opened 2022-08-05
Read existing custom tags from databricks config template

esadler-hbo:
In order to create a new data bricks cluster, we are required to set a series of custom tags. It did not seem like this was directly configurable. 

This was less than 20 lines, so I wanted to share what I did to get it working. It seems like there are a few different ways to do this and I would love to learn of an existing way I missed or contribute a PR to add functionality. 

I will post this in slack and we can discuss there.


xiaoyongzhu (approved):

xiaoyongzhu:
https://feathrai.slack.com/archives/C036HEUQM08/p1659720184974569?thread_ts=1659720184.974569&cid=C036HEUQM08

This thread should have more context

Yuqing-cat (approved):

PR #558, opened 2022-08-07
a possible way to make sure FeathrMvelPluginContext gets serialized and re-loaded on executor nodes

d4ve:
I didn't have time to try this out before going on leave, but wanted to share it with @jaymo001 as a potential fix for an issue we might encounter with the FeathrMvelPluginContext.


xiaoyongzhu:
Close this PR as it is stale for a while. Let me know if anyone wants to reopen it.

PR #560, opened 2022-08-09
Adding workflow to create feathr api/ui image nightly and for each release

jainr:
This workflow builds the docker container for Feathr API/UI and publishes to dockerhub with appropriate tag
 It has two triggers, 
1. nightly i.e. runs every night at 10 pm UTC 
2. Anytime a new branch is created under releases example releases/v0.7.0
 
The final image in dockerhub would have following labels
image


xiaoyongzhu:
per the doc here (https://docs.docker.com/ci-cd/github-actions/), looks like we should use actions such as `docker/build-push-action@v2` rather than pinning it to a specific commit?

jainr:
@xiaoyongzhu - Thanks for the review , removed reference to specific commits and checked that it works.
https://github.com/jainr/feathr/actions/runs/2828455224

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,45 @@
...
+
+on:
+  schedule:
+    # Runs daily at 10 PM UTC, would generate nightly tag
+    - cron: '00 22 * * *'
+  
+  push:
+    # For every push against the releases/** branch, usually would happen at release time, Tag example - releases/v0.7.0
+    branches:
+      - 'releases/**'
```
blrchen, line 16: Looks like current both triggers are configured on release branch. I think nightly image should be generated from main branch?

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          images: feathrfeaturestore/feathr-registry
+      
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          file: FeathrRegistry.Dockerfile
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
```
blrchen, line 44: The docker image tags list should be maintained in this workflow internally. Relay on an external list is not reliable and this would make further maintain hard.

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,45 @@
...
+#   2. Anytime a new branch is created under releases
+
+name: Publish Feathr Docker image to DockerHub
+
+on:
+  schedule:
+    # Runs daily at 10 PM UTC, would generate nightly tag
+    - cron: '00 22 * * *'
+  
+  push:
```
blrchen, line 13: Docker image publish should not get fired for pull request.

.github/workflows/docker-publish.yml
jainr: There is a schedule trigger which runs at 10 pm UTC and it is triggered on main not on release branch, if you don't specify a branch it defaults to main. Currently this builds update two tags, nightly and main.

image

.github/workflows/docker-publish.yml
jainr: While I like the idea, it would make the tags a little bit complex since we are supporting two kind of flows through this pipeline and it would require some custom code to decide nightly tag vs release tag which the  docker/metadata-action@v4 action takes care of.

xiaoyongzhu (approved):

edwincheung (approved):

PR #564, opened 2022-08-09
Optimize purview search logic

enya0405:
## Description

- Improves 'get_project' and similar methods' query logic to reduce time cost 

- The main reason that causes the slowness is we called purview APIs ( such as 'AtlasClient.get_entity') many times, especially in some recursion. It leads to a lot of HTTP requests so that we need to wait for relatively long time for their responses. I found by calling the API 'AtlasClient.get_entity_lineage' we can get all information we need. Then we just need to build edges and entities including their relationships and attributes based on the returned results which costs less time comparing with waiting for many responses from purview.

- Test data:  (tested from backend)
   registry = PurviewRegistry(azure_purview_name = "feathrazuretest3-purview1")
   entity_id_by_name = registry.get_entity_id("enya_test_registry")
   start = time.time()
   project_pre = registry.get_project_origin(entity_id)        // previous 'get_project' API
   print("duration1: ", time.time()-start)
   start = time.time()
   project_curr = registry.get_project(entity_id)                  // current 'get_project' API
   print("duration2: ", time.time()-start)

  duration1:  33.04559826850891
  duration2:  2.3997888565063477

## Dependencies

- [ ] Yes. Make sure to list all the dependencies and licenses.


registry/purview-registry/registry/models.py
```diff
@@ -113,9 +113,18 @@ def __str__(self):
 class RelationshipType(Enum):
     Contains = 1
     BelongsTo = 2
-    Consumes = 3
-    Produces = 4
+    Consumes = 4
```
xiaoyongzhu, line 116: Might need @YihuiGuo 's input here on those changes. I'm fine with it.

xiaoyongzhu:
Thanks @enya0405 . can you update the PR description to reflect:

What was the previous issue that caused it to be slow and what's the improved way?

Also can you put some numbers on the updated logic (like how much time does it take to load the projects)?

registry/purview-registry/registry/models.py
enya0405: I made this change to support the bit operation ' | '. It can make comparison simpler.

YihuiGuo: LGTM, @enya0405 pls make sure you run the test in `registry/purview-registry/test` , espeically `test_creation` and `test_get`.

registry/purview-registry/registry/purview_registry.py
```diff
@@ -137,6 +137,7 @@ def get_all_neighbours(self,id_or_name: Union[str, UUID]) -> list[Edge]:
             relation_lookup[x['attributes']['qualifiedName'].split(self.registry_delimiter)[0]])
              for x in out_edges])
         return result_edges
+
     def get_neighbors(self, id_or_name: Union[str, UUID], relationship: RelationshipType) -> list[Edge]:
```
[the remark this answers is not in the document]
xiaoyongzhu: @enya0405 can double check, but IIRC the search method isn't very scalable and might get results in other projects, and might have performance issues when there are too many entities in the backend. 

In the meanwhile the lineage part is pre calculated, although it has more info than needed, but those info is returned in a batch manner so only a single call is needed.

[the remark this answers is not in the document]
enya0405: @YihuiGuo Do you mean to optimize the 'get_neighbors' at row 141? Replace the 'get_entity_lineage' with some query.

@xiaoyongzhu I will double check the current of returned values and will compare the performance between using get lineage and search query.

registry/purview-registry/registry/models.py
enya0405: Tests done

xiaoyongzhu (approved):

Yuqing-cat (approved):

PR #580, opened 2022-08-13
Generic Input/Output of DataFrames, Client part

windoze:
This PR is the client part corresponding to PR #475 , it contains Python code to generate config/jobparameters to enable the user to use CosmosDb as Data Source and Online/Offline storage.

It adds "CosmosDbSource" and "CosmosDbSink" in the client lib, the former is to use CosmosDb as data source, the latter is to use CosmosDb as online storage.

There are still some open questions about using CosmosDb as offline storage, the main one is, CosmosDb requires every row/doc to have a unique key, but in the feature joining job, we still don't know what should be used as a meaningful key.
For this reason, using CosmosDb as an offline store is not opened in the public interface, even the function is usable internally.


feathr_project/feathr/definition/sink.py
```diff
@@ -100,4 +112,61 @@ def to_feature_config(self) -> str:
...
+        return ret
+
+    def to_argument(self):
+        """
+        One-line JSON string, used by job submitter
+        """
+        return json.dumps(self._to_dict())
+    
+class CosmosDbSink(GenericSink):
+    def __init__(self, name: str, endpoint: str, database: str, collection: str):
```
blrchen, line 159: should change collection to container to reduce confusion

feathr_project/feathr/definition/sink.py
```diff
@@ -1,12 +1,18 @@
-from typing import List, Optional
+from abc import abstractmethod
+import json
+from typing import Dict, List, Optional
 from jinja2 import Template
+from pymysql import paramstyle
```
blrchen, line 5: Remove pymysql as current code does not use it

windoze: Removed

feathr_project/feathr/definition/sink.py
windoze: Changed

blrchen (approved):

blrchen:
E2E test should get updated to cover write features to cosmos db

feathr_project/feathr/client.py
```diff
@@ -59,9 +78,10 @@ class FeatureGenerationJobParams:
         feature_config: Path to the features config.
     """
 
-    def __init__(self, generation_config_path, feature_config):
+    def __init__(self, generation_config_path, feature_config, secrets=[]):
```
xiaoyongzhu, line 81: Add docstring for how this secrets will be used

feathr_project/feathr/client.py
```diff
@@ -579,16 +605,16 @@ def materialize_features(self, settings: MaterializationSettings, execution_conf
                 raise RuntimeError("Please call FeathrClient.build_features() first in order to materialize the features")
 
             udf_files = _PreprocessingPyudfManager.prepare_pyspark_udf_files(settings.feature_names, self.local_workspace_dir)
-            # CLI will directly call this so the experience won't be broken
-            self._materialize_features_with_config(config_file_path, execution_configurations, udf_files)
+            # CLI will directly call this so the experiene won't be broken
```
xiaoyongzhu, line 608: experience

feathr_project/feathr/definition/sink.py
```diff
@@ -100,4 +111,61 @@ def to_feature_config(self) -> str:
...
         hocon_config = tm.render(sink=self)
-        return hocon_config
\ No newline at end of file
+        return hocon_config
+
+    def to_argument(self):
+        return self.output_path
+
+    
+class GenericSink(Sink):
```
xiaoyongzhu, line 120: Add docstring for this. What is the purpose for `GenericSink`?

feathr_project/feathr/definition/source.py
```diff
@@ -215,5 +251,93 @@ def to_feature_config(self) -> str:
         msg = tm.render(source=self, brokers=brokers, topics=topics)
         return msg
 
+    def to_argument(self):
+        raise TypeError("KafKaSource cannot be used as observation source")
+
+class GenericSource(Source):
```
xiaoyongzhu, line 257: Same comment for GenericSource, should we make it abstract?

jainr:
Can we add test cases to show this in working?

blrchen (approved):

Yuqing-cat (dismissed):

feathr_project/feathr/definition/sink.py
windoze: It's the base class of CosmosDbSink, while itself can be used directly, not meant to end users but for other codes, like we plan to add job submission function in the registry.

feathr_project/feathr/client.py
windoze: Added

feathr_project/feathr/definition/sink.py
xiaoyongzhu: Can we add those info in the code comments for easier maintenance?

docs/how-to-guides/jdbc-cosmos-notes.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+* `name` is the source name, same as other sources.
+* `url` is the database URL in JDBC format.
+* `dbtable` or `query`, `dbtable` is the table name in the database and `query` is a SQL `SELECT` statement, only one of them should be specified at the same time.
+* `auth` can be `None`, `'USERPASS'` or `'TOKEN'`.
+
+There are some other parameters such as `preprocessing`, they're same as other sources like `HdfsSource`.
+
+After creating the `JdbcSource`, you can use it in the same way as other kinds of sources.
```
xiaoyongzhu, line 28: provide code example or link to test cases so end users know the format

feathr_project/test/test_azure_spark_e2e.py
```diff
@@ -201,3 +202,73 @@ def test_feathr_get_offline_features():
...
+                                    output_path=output_path)
+
+        # assuming the job can successfully run; otherwise it will throw exception
+        client.wait_job_to_finish(timeout_sec=Constants.SPARK_JOB_TIMEOUT_SECONDS)
+
+        # We may not have SQL client package installed so skip the verification
+        # res_df = get_result_df(client)
+        # assert res_df.shape[0] > 0
+
+@pytest.mark.skip(reason="We don't have CosmosDb configured in CI pipeline")
```
xiaoyongzhu, line 250: Let's make sure we configure those resources in the test environment and make sure those test can run successfully

feathr_project/test/test_azure_spark_e2e.py
```diff
@@ -201,3 +202,73 @@ def test_feathr_get_offline_features():
...
+    collection = ''.join(['feathrazure_cijob_materialize_cosmosdb_','_', str(now.minute), '_', str(now.second), ""])
+    sink = CosmosDbSink(name='o', endpoint='https://feathrtestcosmosdb.documents.azure.com:443/', database='feathr', collection=collection)
+    settings = MaterializationSettings("nycTaxiTable",
+                                       sinks=[sink],
+                                       feature_names=[
+                                           "f_location_avg_fare", "f_location_max_fare"],
+                                       backfill_time=backfill_time)
+    client.materialize_features(settings)
+    # assuming the job can successfully run; otherwise it will throw exception
+    client.wait_job_to_finish(timeout_sec=Constants.SPARK_JOB_TIMEOUT_SECONDS)
```
xiaoyongzhu, line 274: How to verify if the values are actually materialized? We should add command to get those values and check them (for example, not NULL)

windoze: Added in doc, will update this test later.

feathr_project/test/test_azure_spark_e2e.py
xiaoyongzhu: Let's talk about it today. We should have the cloud test setup rather than delaying it.

feathr_project/feathr/definition/sink.py
windoze: Comments added

feathr_project/test/test_azure_spark_e2e.py
xiaoyongzhu: BTW it should only take less than 1 hour to setup the cloud resource and make sure all the testing can pass (if they have been verified)

xiaoyongzhu (approved):

jainr (approved):

xiaoyongzhu:
The test failures are irrelevant hence merging the code.

PR #581, opened 2022-08-16
Misc issue fixes

jaymo001:
## Description
This PR fixes multiple minor issues related to Feathr backward compatibility.

1. Support Tensor in Header by replacing FeatureTypes with FeatureTypeConfig.
2. Fix a bug related to derived feature function adaptor in config loader, where the adapted derived feature function is not being used.
3. Fix a bug related to keyExtractor where keyExpression is null.
4. Add test cases

No.

### DIFF
--- a/src/main/scala/com/linkedin/feathr/offline/util/FeaturizedDatasetMetadata.scala
+++ b/src/main/scala/com/linkedin/feathr/offline/util/FeaturizedDatasetMetadata.scala
@@ -1,4 +1,10 @@
 package com.linkedin.feathr.offline.util
 
-case class FeaturizedDatasetMetadata(meta: Map[String, String] = Map()) {
+import com.linkedin.feathr.common.Header
+/**
+ * The metadata for FeaturizedDataset
+ * @param meta extra metadata
+ * @param header feature type header info
+ */
+case class FeaturizedDatasetMetadata(meta: Map[String, String] = Map(), header: Option[Header] = None ) {
 }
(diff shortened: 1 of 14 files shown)


xiaoyongzhu (approved):

src/test/scala/com/linkedin/feathr/offline/TestFeathrUdfPlugins.scala
```diff
@@ -79,39 +99,39 @@ class TestFeathrUdfPlugins extends FeathrIntegTest {
...
-//            // a_id
-//            "3",
-//            // featureWithNull
-//            3.0f))),
-//      StructType(
-//        List(
-//          StructField("a_id", StringType, true),
-//          StructField("featureWithNull", FloatType, true))))
-//    def cmpFunc(row: Row): String = row.get(0).toString
-//    FeathrTestUtils.assertDataFrameApproximatelyEquals(filteredDf, expectedDf, cmpFunc)
```
hangfei, line 115: remove this?

hangfei (approved):

src/main/scala/com/linkedin/feathr/offline/util/FeaturizedDatasetMetadata.scala
```diff
@@ -1,4 +1,10 @@
 package com.linkedin.feathr.offline.util
 
-case class FeaturizedDatasetMetadata(meta: Map[String, String] = Map()) {
+import com.linkedin.feathr.common.Header
+/**
+ * The metadata for FeaturizedDataset
+ * @param meta extra metadata
+ * @param header feature type header info
+ */
+case class FeaturizedDatasetMetadata(meta: Map[String, String] = Map(), header: Option[Header] = None ) {
```
d4ve, line 9: Should be more docs about what this means; what a featurized-dataset is? (Do we have docs about it in this public version of feathr? Should we just call it as feathr's output dataset that includes info about the tensor types of joined features...?)

Also, what does "extra metadata" mean; what kinds of things is the map allowed/expected to contain, what are the keys, values, etc.

PR #588, opened 2022-08-17
Remove extra dependencies to avoid dead loop of page re-rendering

enya0405:
## Description

- Remove extra dependencies for 'useEffect' and 'useCallback'. These dependencies can cause page re-rendering in a dead loop and infinite console errors would be printed.
- Add the eslint disable mark to avoid warnings from eslint for these lines 

### DIFF
--- a/ui/src/components/graph/graph.tsx
+++ b/ui/src/components/graph/graph.tsx
@@ -37,7 +37,7 @@ const Graph = ({ data, nodeId }: Props) => {
 
   useEffect(() => {
     setElements(layoutedElements);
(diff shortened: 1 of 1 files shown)


blrchen (approved):

jainr (commented):
Not familiar with eslint but why are we disabling it?

enya0405:
> 

It requires us setting some variables as dependencies. Some of them are not necessary and can cause infinite page re-rending in a dead loop. So the console will keep printing that error message which may cause serious problem.

xiaoyongzhu (approved):

PR #593, opened 2022-08-19
Fix graph edge errors when calling SQL APIs

enya0405:
## Description

- Fix some bugs in the sql registry API "get_project". These bugs can cause lost of some necessary edges among entities.


registry/sql-registry/registry/db_registry.py
```diff
@@ -405,16 +405,18 @@ def _fill_entity(self, e: Entity) -> Entity:
         return e
 
     def _get_edges(self, ids: list[UUID], types: list[RelationshipType] = []) -> list[Edge]:
-        if not ids or not types:
+        if not ids:
             return []
         sql = fr"""select edge_id, from_id, to_id, conn_type from edges
         where from_id in %(ids)s
         and to_id in %(ids)s"""
+        types = [ str(t) for t in types]
```
windoze, line 413: Duplicated with line 422.
Also should use `name` property for `Enum` objects.

registry/sql-registry/registry/models.py
```diff
@@ -117,6 +117,15 @@ class RelationshipType(Enum):
     Produces = 4
 
 
+    def __str__(self):
```
windoze, line 120: This actually already exists as `Enum.name` property.

registry/sql-registry/registry/db_registry.py
```diff
@@ -405,16 +405,18 @@ def _fill_entity(self, e: Entity) -> Entity:
         return e
 
     def _get_edges(self, ids: list[UUID], types: list[RelationshipType] = []) -> list[Edge]:
-        if not ids or not types:
+        if not ids:
```
windoze, line 408: Shouldn't we just return an empty list if no input `types` is specified?

enya0405: Based on the line 94 where this method is called "all_edges = self._get_edges(ids)", and the sql query on line 410 when the types is empty, my understanding is when types is empty we should query for results containing all relationship types. Otherwise, the 'all_edges' will return empty list which is the root cause of this error. We can also set all relationship types at line 94, but I think it will be less efficient.

registry/sql-registry/registry/db_registry.py
enya0405: Added for debugging. Will remove it.

registry/sql-registry/registry/models.py
enya0405: Added this one to support str(t) at line 422. Or to modify it to t.name as you mentioned.

windoze (approved):

blrchen (approved):

PR #600, opened 2022-08-23
Fix source entity missing time window bug.

YihuiGuo:
The bug is about source entity missing time window information. turns out timestamp column and timestamp format. 
These properties are lost when saving to purview.
Add these columns back, to avoid breaking other pipeline, do not remove these camel case properties.


xiaoyongzhu (approved):

windoze (approved):

jainr:
Is camel casing required for backward compatibility?

PR #602, opened 2022-08-24
Decode log error message

enya0405:
## Description

- Decode log error messages which begin with the char 'b' as the mark of byte string.s So that escape chars within messages can work well and can be printed out by a clear format.
- Error messages sample:

(Before the change)
Screen Shot 2022-08-24 at 16 24 10
(After the change)
Screen Shot 2022-08-24 at 16 24 49


xiaoyongzhu (approved):

xiaoyongzhu:
maybe paste the error message in the pr description to make it clear for other approvers.

enya0405:
> maybe paste the error message in the pr description to make it clear for other approvers on why this need to be changed.

Thanks for the suggestion. Added some screenshot to show the difference.

blrchen (approved):

PR #603, opened 2022-08-24
Use maven based submission by default for samples

xiaoyongzhu:
## Description

This PR has a few improvements:

1. Use Maven jar version 0.6.2 for submission
2. delete all the `feathr_runtime_location` in the samples so that users don't use them in the future. By removing the `feathr_runtime_location` config, Feathr will automatically using jars from maven.
3. Format the notebook to JSON to make it easier for future review


blrchen (approved):

jainr:
How does the notebook get the jar from maven?

jainr (approved):

xiaoyongzhu:
> 2\. By removing the `feathr_runtime_location` config, Feathr will automatically using jars from maven.

Yeah if the `feathr_runtime_location` config is not set, Feathr will automatically using jars from maven.

PR #618, opened 2022-08-26
Print lineage endpoint link in log after registering features

enya-yx:
## Description

- Print out link to check project lineage after registering features in log so that customers can check it conveniently
- The log would look like:
   2022-08-26 16:12:04.175 | INFO     | feathr.registry._feathr_registry_client:register_features:126 - Check project lineage 
   by this link: https://feathr-purview-registry.azurewebsites.net/projects/registry_test_enya_3/lineage

### DIFF
--- a/feathr_project/feathr/registry/_feathr_registry_client.py
+++ b/feathr_project/feathr/registry/_feathr_registry_client.py
@@ -2,6 +2,7 @@ import importlib
 import inspect
 import json
 import logging
+from loguru import logger
 import os
 from pathlib import Path
 import sys
@@ -25,7 +26,6 @@ from feathr.definition.transformation import ExpressionTransformation, Transform
 from feathr.definition.typed_key import TypedKey
 from feathr.registry.feature_registry import FeathrRegistry
 from feathr.utils._file_utils import write_to_file
-
 
 def to_camel(s):
     if not s:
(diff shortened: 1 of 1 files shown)


feathr_project/feathr/registry/_feathr_registry_client.py
```diff
@@ -122,6 +122,8 @@ def register_features(self, workspace_path: Optional[Path] = None, from_context:
         for df in _topological_sort(derived_feature_list):
             if not hasattr(df, "_registry_id"):
                 df._registry_id = self._create_derived_feature(df)
+        url = '/'.join(self.endpoint.split('/')[:3])       
+        logger.info(f"Check project lineage by this link: {url}/projects/{self.project_name}/lineage")
```
blrchen, line 126: If user register with direct purview write, which means no endpoint is specified in config, will this code work?

enya-yx: For that case, this method will not be called. The 'register_features' method in '_feathr_registry_purview' will be called instead with the given purview name.

blrchen (approved):

feathr_project/feathr/registry/_feathr_registry_client.py
Yuqing-cat: Just to confirm, do we have a similar log after registering features in the direct purview write mode?

Yuqing-cat (approved):

PR #622, opened 2022-08-28
Windoze/elasticsearch

windoze:
This PR enables using ElasticSearch as the online and offline store.

The CI test has been added but is currently marked as skipped because we need to setup ES cluster in our testing environment first.

The PR also enables using ElasticSearch as a source but I don't see any strong need for this usage, we can either hide it from the end users or remove this part.

### DIFF
--- a/docs/how-to-guides/jdbc-cosmos-notes.md
+++ b/docs/how-to-guides/jdbc-cosmos-notes.md
@@ -1,6 +1,6 @@
 ---
 layout: default
-title: Using SQL databases and CosmosDb with Feathr
+title: Using SQL databases, CosmosDb, and ElasticSearch with Feathr
 parent: Feathr How-to Guides
 ---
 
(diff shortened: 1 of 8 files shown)


hangfei:
Thanks for the PR!

Could you explain the preferred use case for ES?

and I don't quite understand: `The PR also enables using ElasticSearch as a source but I don't see any strong need for this usage`. Could you elaborate? Thanks.

windoze:
> Thanks for the PR!
> 
> Could you explain the preferred use case for ES?

ES has good performance/scalability and supports composited types, a typical case is to use ES as the online store just like Redis, without the trouble of encoding/decoding.

> and I don't quite understand: `The PR also enables using ElasticSearch as a source but I don't see any strong need for this usage`. Could you elaborate? Thanks.

People can use ELK to store logs, so it's possible that some users need to extract features from ES, so being able to use ES as the data source could be helpful sometimes. But so far I don't see our customers asking for this feature.

src/main/scala/com/linkedin/feathr/swj/transformer/FeatureTransformer.scala
```diff
@@ -93,6 +93,7 @@ object FeatureTransformer {
     // The order of the feature columns should follow the order of the specified sliding window
     // features. In addition, the feature columns should be in front of join key and timestamp
     // columns. These will guarantee we can fetch corresponding feature columns using index.
+    println(s"SQL: ${sql}")
```
xiaoyongzhu, line 96: should we remove this print?

xiaoyongzhu:
This LGTM. Although ES is not asked a lot, but it's a feature that's nice to have and we can continue improving it once there's someone using it.

docs/how-to-guides/jdbc-cosmos-notes.md
```diff
@@ -102,4 +102,46 @@ db_client = client.get_database_client(some_cosmosdb_database)
 container_client = db_client.get_container_client(some_cosmosdb_collection)
 doc = container_client.read_item(some_key)
 feature_value = doc['feature_name']
-```
\ No newline at end of file
+```
+
+## Using ElasticSearch as online store
```
xiaoyongzhu, line 107: Maybe we should put a "Beta" or "preview" here to lower the expectation?

windoze: This should have the same quality as CosmosDb, they share exactly the same underlying code, the only difference is the user interaction part in the Feathr client, which is merely a wrapper.

src/main/scala/com/linkedin/feathr/swj/transformer/FeatureTransformer.scala
windoze: Let's keep it as-is, this is helpful for diagnostic, especially when the issue happens on the customer site.
Also, it's not a security leak, no info more sensitive than the feature definitions will be exposed in the output.

xiaoyongzhu: Talked offline, shall we call it something like "Generated SQL statement" so that it's clear for debug purpose?

xiaoyongzhu (approved):

src/main/scala/com/linkedin/feathr/swj/transformer/FeatureTransformer.scala
windoze: Fixed

Yuqing-cat (approved):

PR #657, opened 2022-09-09
Adding DevSkim linter to Github actions

jainr:
## Description
Adding DevSkim Plugin to Github Action

DevSkim is security linter that highlights common security issues in source code.

The DevSkim GitHub Action outputs a sarif file compatible with GitHub's Security Issues view.

It is strongly suggested to be run for OSS projects.

## How was this PR tested?
Tested in my fork, here is a successful run.

### DIFF
--- a/.github/workflows/devskim-security-linter.yml
+++ b/.github/workflows/devskim-security-linter.yml
@@ -1,0 +1,35 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party (Microsoft) and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+# For more details about Devskim, visit https://github.com/marketplace/actions/devskim 
+
+name: DevSkim
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+  schedule:
+    - cron: '25 4 * * 2'
+
+jobs:
+  lint:
+    name: DevSkim
+    runs-on: ubuntu-20.04
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+    steps:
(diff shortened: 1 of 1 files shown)


blrchen:
Can you please take a look at the failures as well? Either fix them or exclude them if it's false alarm. Then we can have a baseline with DevSkim pass on main branch after this is merged.

xiaoyongzhu:
@jainr thanks for the PR, can you also help fix some of the alerts that this tool is giving? Some of them might be false alarms and some of them might need to be fixed.

jainr:
The current two alerts come from documentation, and it is about https URLs. The first one is **curl -s https://whatismyip.akamai.com**/ If I change the URL to https, it won't work. I don't want to skip the folders completely as it might skip some legit issues.

The second URL it complains about is     **export RBAC_REGISTRY_URL="http://127.0.0.1:18000/api/v1"**. It is a localhost URL and in order for us to switch over to https we would have to use TLS certificate.

It's safe to ignore these two alerts, but we should not skip the folders for any future issues.

Yuqing-cat:
LGTM, but if it's merged does it means that we will always get failure in DevSkim and have to check if there are a third error?

jainr:
> LGTM, but if it's merged does it means that we will always get failure in DevSkim and have to check if there are a third error?

Devskim does not fail, at least from GH action perspective and it does not print the reason in the execution logs either, since it is sensitive information. It provides the findings under Security tab so yes we have to monitor it from time to time.

image


Here is how it looks under GH action.
https://github.com/jainr/feathr/actions/workflows/devskim-security-linter.yml

edwincheung (approved):

Yuqing-cat (approved):

PR #674, opened 2022-09-15
Enhance SQL Registry Error Messages

windoze:
## Description

Right now the SQL registry returns all errors as `500 Internal Error`, this PR improves the error handling and returns 400/404/409 on corresponding criteria.

Also it introduces an environment variable `REGISTRY_DEBUGGING`, the returned HTTP error will include the detailed track back info when it's set to non-empty string. This variable should only be used for debugging purposes.

### DIFF
--- a/registry/sql-registry/main.py
+++ b/registry/sql-registry/main.py
@@ -37,7 +37,7 @@ def exc_to_content(e: Exception) -> dict:
     return content
 
 @app.exception_handler(ConflictError)
-async def http_exception_handler(_, exc: ConflictError):
+async def conflict_error_handler(_, exc: ConflictError):
     return JSONResponse(
         status_code=409,
         content=exc_to_content(exc),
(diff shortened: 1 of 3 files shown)


jainr (commented):
Can we add some test cases or update existing test cases to show how new error codes are thrown?

xiaoyongzhu:
lgtm. Can we also add the env in the docs?

blrchen:
Can we add same on purview as well? I remember most of the issues seen recently happened on purview registry.

xiaoyongzhu:
> Can we add same on purview as well? I remember most of the issues seen recently happened on purview registry.

Agree, most of the problems seems to be in Purview backend.

Yuqing-cat (approved):
LGTM. Will Purview Registry follow the same Error Message pattern?

blrchen (approved):
Approved, I will have a different PR to port the fix to Purview registry.

PR #676, opened 2022-09-16
Fix multi-keyed feature in anchor (direct purview)

YihuiGuo:
When using old purview client (not registry client), if features inside an anchor has a same SET of keys ,  when calling get_feature_from_registry, only the first key will be used. 

This PR handles the situation where each feature has multiple keys, and the collection of keys are identical among features inside an anchor.


Yuqing-cat (approved):

blrchen (approved):
Approve to unblock end user. 

Yihiui agree to add test case as following up PR.

PR #682, opened 2022-09-19
Switch purview test endpoint

YihuiGuo:
## Description

Resolves #XXX

## How was this PR tested?
Need to be marked as "safe to test", then monitor the test result


xiaoyongzhu:
Should we do both instead of just purview?

Yuqing-cat:
@blrchen is refactoring the test case right now.

PR #684, opened 2022-09-19
Fix path with #LATEST

jaymo001:
## Description
Fix a bug in resolve latest where the path with #LATEST  is not resolved correctly or used as expected.

Root cause: #LATEST in source.path is resolved but location.path is not.

This is mainly for backward compatibility.

## How was this PR tested?

Added unit test

No


d4ve (commented):
Is the behavior for `#LATEST` in the docs somewhere? Can you offer any more detail on what the bug was (summary of incorrect behavior -> correct behavior)?

jaymo001:
> Is the behavior for `#LATEST` in the docs somewhere? Can you offer any more detail on what the bug was (summary of incorrect behavior -> correct behavior)?

Updated description. This is mainly for backward compatibility.

xiaoyongzhu (approved):

d4ve (approved):

PR #690, opened 2022-09-21
Lock python dependency versions

xiaoyongzhu:
## Description

Currently many of the python package versions are not locked which caused quite some troubles during installation and we've seen multiple issues there. This PR aims to lock all the python dependencies to the latest tested versions. Also solves https://github.com/feathr-ai/feathr/issues/687


Yuqing-cat (dismissed):

feathr_project/setup.py
```diff
@@ -20,35 +20,37 @@
...
+        "graphlib_backport<=1.0.3",
         "protobuf==3.*",
-        "azure-keyvault-secrets",
-        "confluent-kafka",
-        "databricks-cli",
-        "avro",
+        "confluent-kafka<=1.9.2",
+        "databricks-cli<=0.17.3",
+        "avro<=1.11.1",
+        "azure-storage-file-datalake>=12.5.0",
```
windoze, line 45: I tried 12.8.0 but it doesn't seem to work, maybe we need a more strict constraint here.

xiaoyongzhu: good point

windoze (approved):

blrchen (approved):

PR #691, opened 2022-09-21
Allow recreating entities for PurView registry

### LINKED ISSUE #675: [FR] Support Entity Updates (project, feature etc) in Purview Registry
### Willingness to contribute

### Feature Request Proposal

Currently for Purview backend updates are not allowed to entities.
This means customers/users can not register any features to a project (once they have already registered some) or update existing features.

Allow for updating an existing entity

### Motivation

Currently for Purview backend updates are not allowed to entities.
This means customers/users can not register any features to a project (once they have already registered some) or update existing features.

Allow for updating an existing entity

### Details

_No response_

windoze:
## Description

Resolves #675 partly, we still do not allow updating on any existing features and sources, but we do allow re-creating features and sources as long as the definitions keep the same, and we allow adding new entities into existing containers such as Anchor and Project.

### DIFF
--- a/feathr_project/feathr/registry/_feature_registry_purview.py
+++ b/feathr_project/feathr/registry/_feature_registry_purview.py
@@ -750,19 +750,19 @@ derivations: {
             if j["typeName"] == response["typeName"]:
                 if j["typeName"] == "Process":
                     if response["attributes"]["qualifiedName"] != j["attributes"]["qualifiedName"]:
-                        raise RuntimeError("The requested entity conflicts with the existing entity in PurView")
+                        raise RuntimeError("The requested entity %s conflicts with the existing entity in PurView" % j["attributes"]["qualifiedName"])
                 else:
-                    if "type" in response['attributes'] and response["typeName"] in ("feathr_anchor_feature_v1", "feathr_derived_feature_v1"):
+                    if "type" in response['attributes'] and response["typeName"] in (TYPEDEF_ANCHOR_FEATURE, TYPEDEF_DERIVED_FEATURE):
                         conf = ConfigFactory.parse_string(response['attributes']['type'])
                         response['attributes']['type'] = dict(conf)
                     keys = set([_to_snake(key) for key in j["attributes"].keys()]) - set(["qualified_name"])
                     keys.add("qualifiedName")
                     for k in keys:
                         if response["attributes"][k] != j["attributes"][k]:
-                            raise RuntimeError("The requested entity conflicts with the existing entity in PurView")
+                            raise RuntimeError("The requested entity %s conflicts with the existing entity in PurView" % j["attributes"]["qualifiedName"])
                 return response["guid"]
             else:
-                raise RuntimeError("The requested entity conflicts with the existing entity in PurView")
+                raise RuntimeError("The requested entity %s conflicts with the existing entity in PurView" % j["attributes"]["qualifiedName"])
         except AtlasException as e:
             pass
         
--- a/feathr_project/feathr/registry/_feature_registry_purview.py
+++ b/feathr_project/feathr/registry/_feature_registry_purview.py
@@ -3,6 +3,7 @@ import importlib
 import inspect
 import itertools
 import os
+import re
 import sys
 import ast
 import types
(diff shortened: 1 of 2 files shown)


feathr_project/feathr/registry/_feature_registry_purview.py
```diff
@@ -720,6 +740,32 @@ def upload_single_entity_to_purview(self,entity:Union[AtlasEntity,AtlasProcess])
...
+                    keys = set([_to_snake(key) for key in j["attributes"].keys()]) - set(["qualified_name"])
+                    keys.add("qualifiedName")
+                    for k in keys:
+                        if response["attributes"][k] != j["attributes"][k]:
+                            raise RuntimeError("The requested entity conflicts with the existing entity in PurView")
+                return response["guid"]
+            else:
+                raise RuntimeError("The requested entity conflicts with the existing entity in PurView")
+        except AtlasException as e:
+            pass
```
xiaoyongzhu, line 767: Should we throw out the exception and print out, rather than pass?

registry/purview-registry/registry/purview_registry.py
```diff
@@ -568,17 +571,47 @@ def _register_feathr_feature_types(self):
...
+            """
+            Try to find existing entity/process first, if found, return the existing entity's GUID
+            """
+            id = self.get_entity_id(entity.qualifiedName)
+            response =  self.purview_client.get_entity(id)['entities'][0]
+            j = entity.to_json()
+            if j["typeName"] == response["typeName"]:
+                if j["typeName"] == "Process":
+                    if response["attributes"]["qualifiedName"] != j["attributes"]["qualifiedName"]:
+                        raise RuntimeError("The requested entity conflicts with the existing entity in PurView")
```
xiaoyongzhu, line 589: Also illustrate the request entity name. Something like below. Also PurView should be Purview:

The requested entity {entityName} conflicts with the existing entity in Purview.

registry/purview-registry/registry/purview_registry.py
```diff
@@ -568,17 +571,47 @@ def _register_feathr_feature_types(self):
...
+                else:
+                    if "type" in response['attributes'] and response["typeName"] in ("feathr_anchor_feature_v1", "feathr_derived_feature_v1"):
```
xiaoyongzhu, line 591: use the constant name rather than hard coded string `feathr_anchor_feature_v1`

registry/purview-registry/registry/purview_registry.py
```diff
@@ -568,17 +571,47 @@ def _register_feathr_feature_types(self):
...
+                        conf = ConfigFactory.parse_string(response['attributes']['type'])
+                        response['attributes']['type'] = dict(conf)
+                    keys = set([to_snake(key) for key in j["attributes"].keys()]) - set(["qualified_name"])
```
xiaoyongzhu, line 594: Can you explain a bit what you are trying to do here for the set operation? Might be good to add some comments for easier maintenance.

registry/purview-registry/registry/purview_registry.py
```diff
@@ -568,17 +571,47 @@ def _register_feathr_feature_types(self):
...
+                    keys.add("qualifiedName")
+                    for k in keys:
+                        if response["attributes"][k] != j["attributes"][k]:
+                            raise RuntimeError("The requested entity conflicts with the existing entity in PurView")
+                entity.guid = response["guid"]
+                return
             else:
-                raise RuntimeError("Feature registration failed.", results)            
+                raise RuntimeError("The requested entity conflicts with the existing entity in PurView")
```
xiaoyongzhu, line 602: Same here on the error message. We should make it more illustrative.


The requested entity {entityName} conflicts with the existing entity in Purview.

registry/purview-registry/registry/purview_registry.py
```diff
@@ -568,17 +571,47 @@ def _register_feathr_feature_types(self):
...
+        except AtlasException as e:
+            pass
```
xiaoyongzhu, line 604: Should we print out the exception rather than simply pass?

windoze: This `pass` is deliberated, Atlas will throw an exception when the entity is not found, this code piece uses this behavior to check entity existence.

feathr_project/feathr/registry/_feature_registry_purview.py
windoze: This pass is deliberated, Atlas will throw an exception when the entity is not found, this code piece uses this behavior to check entity existence.

registry/purview-registry/registry/purview_registry.py
windoze: Fixed

registry/purview-registry/registry/purview_registry.py
windoze: @YihuiGuo has some fix that generates duplicated keys such as `timestamp_column_name` and `timestampColumnName` both exist in the requested entity.
I already asked her to fix the issue, but removing this logic must happen after her fix.

registry/purview-registry/registry/purview_registry.py
windoze: Fixed

registry/purview-registry/registry/purview_registry.py
windoze: Fixed

xiaoyongzhu (approved):

edwincheung (approved):

jainr:
@windoze - From end user perspective, if I run the sample notebook two times against Purview backend, it should work, since feature definition will remain same?

Can you add more examples of what is allowed and what is not in terms of Purview updates?

windoze:
> @windoze - From end user perspective, if I run the sample notebook two times against Purview backend, it should work, since feature definition will remain same?
> 
> Can you add more examples of what is allowed and what is not in terms of Purview updates?

The current design is: the recreating should succeed as long as **all attributes** remain exactly the same as the old one, and the returned id also remains the same. All other cases will trigger a 402/409 conflict error. Please report if you see any other behavior as they should be bugs.

We added this constraint for future compatibility since we still didn't figure out what should be the most appropriate behavior to update the existing entities.

PR #709, opened 2022-09-27
Enhance purview registry error messages

blrchen:
## Description

This PR is to follow up on PR #674 to port the fix to purview registry.

## How was this PR tested?

Test register same project twice with python client to purview registry, the 2nd registration now fails with 409 conflict error code: 
```
RuntimeError: Failed to call registry API, status is 409, error is {"message":"Entity -1001, feathr_workspace_v1 -- blair_0927.1 already exists in Purview. Please use a new name."}
```

### DIFF
--- a/registry/purview-registry/main.py
+++ b/registry/purview-registry/main.py
@@ -1,11 +1,12 @@
 import os
+import traceback
 from re import sub
 from typing import Optional
 from uuid import UUID
 from fastapi import APIRouter, FastAPI, HTTPException
+from fastapi.responses import JSONResponse
 from starlette.middleware.cors import CORSMiddleware
-from registry import *
-from registry.purview_registry import PurviewRegistry
+from registry.purview_registry import PurviewRegistry, ConflictError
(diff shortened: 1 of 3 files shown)


registry/purview-registry/main.py
```diff
@@ -43,6 +44,48 @@ def to_camel(s):
                    allow_headers=["*"],
                    )
 
+def exc_to_content(e: Exception) -> dict:
+    content={"message": str(e)}
+    if os.environ.get("REGISTRY_DEBUGGING"):
```
xiaoyongzhu, line 49: Should we document all the envs somewhere?

jainr (approved):
LGTM

registry/purview-registry/main.py
blrchen: Good catch, doc is updated

Yuqing-cat (approved):

xiaoyongzhu (approved):

PR #711, opened 2022-09-27
Fix protobuf version

enya-yx:
## Description

- Fix protobuf version to avoid incompatibility caused by versions


feathr_project/setup.py
```diff
@@ -38,7 +38,7 @@
         # fixing https://github.com/feathr-ai/feathr/issues/687
         "deltalake<=0.5.8",
         "graphlib_backport<=1.0.3",
-        "protobuf==3.*",
+        "protobuf==3.19.4",
```
xiaoyongzhu, line 41: Can we have something like >= or <= rather than fixing it to a certain version?

enya-yx: We cannot make sure if all versions above or below a certain version can work for now.

xiaoyongzhu: I don't think we have hard requirements for the protobuf 3 version (3.19.4), since we are just using it generically. Having this the strict version will be hard to run in multiple notebook services like synapse/databricks.

blrchen:
Since the main reason of this PR is protobuf==3.20.2 has issue, might be we should do <=3.20.1 instead?

We should also add a note for why 3.20.1 is used by including https://github.com/protocolbuffers/protobuf/issues/10571 in comemnt.

enya-yx:
> Since the main reason of this PR is protobuf==3.20.2 has issue, might be we should do <=3.20.1 instead?
> 
> We should also add a note for why 3.20.1 is used by including [protocolbuffers/protobuf#10571](https://github.com/protocolbuffers/protobuf/issues/10571) in comemnt.

Do we need to do some research to make sure all protobuf versions <= 3.20.1 can work well? Or this is just a suggestion?

jainr:
> > Since the main reason of this PR is protobuf==3.20.2 has issue, might be we should do <=3.20.1 instead?
> > We should also add a note for why 3.20.1 is used by including [protocolbuffers/protobuf#10571](https://github.com/protocolbuffers/protobuf/issues/10571) in comemnt.
> 
> Do we need to do some research to make sure all protobuf versions <= 3.20.1 can work well? Or this is just a suggestion?

The way I understand protobuf<=3.19.4 will work is it will install 3.19.4 version or the next lower version if it is not available.

windoze:
Is 3.20.2 the only problematic version? Or all versions above have the issue?

blrchen (approved):

blrchen:
Btw, the issue you reported on protobuf 3.20.2 has been fixed. See https://github.com/protocolbuffers/protobuf/issues/10571 for details

Issue #715, opened 2022-09-29
[FR] `__version__` to feathr


loomlike:
### Willingness to contribute

Yes. I can contribute a fix for this bug independently.

### Feature Request Proposal

Add version info to python sdk.

### Motivation

> #### What is the use case for this feature?
With `feathr.__version__`, it will be easy to check which version I'm using without looking into setup.py or installed pip packages.

### What component(s) does this feature request affect?

- [X] `Python Client`: This is the client users use to interact with most of our API. Mostly written in Python.
- [ ] `Computation Engine`: The computation engine that execute the actual feature join and generation work. Mostly in Scala and Spark.
- [ ] `Feature Registry API`: The frontend API layer supports SQL, Purview(Atlas) as storage. The API layer is in Python(FAST API)
- [ ] `Feature Registry Web UI`: The Web UI for feature registry. Written in React

PR #725, opened 2022-10-05
Fix function string parsing

### LINKED ISSUE #724: [BUG] preprocessing functions parsing error when load the features from registry
### Willingness to contribute

### Feathr version

0.8.0

### System information

- **OS Platform and Distribution (e.g., Linux Ubuntu 20.0)**: Linux Ubuntu 20
- **Python version**: 3.10

### Describe the problem

`_PreprocessingPyudfManager._parse_function_str_for_name` has a bug so that if the udf function includes type hint, it parses wrong function name.

### Tracking information

_No response_

### Code to reproduce bug

```
def preprocessing(df: DataFrame) -> DataFrame:
    import pyspark.sql.functions as F
    df = df.withColumn("fare_amount_cents", (F.col("fare_amount") * 100.0).cast("float"))
    return df
```

then run 
```
client.register_features()
client.get_features_from_registry()
```

loomlike:

## Description

Resolves #724 #715

## How was this PR tested?

via unit test -- This commit also adds the unit test code.
Also tested manually by registering and loading the features to/from purview.

### DIFF
--- a/feathr_project/feathr/udf/_preprocessing_pyudf_manager.py
+++ b/feathr_project/feathr/udf/_preprocessing_pyudf_manager.py
@@ -1,12 +1,15 @@
+import ast
 import inspect
 import os
+import pickle
 from pathlib import Path
 from typing import List, Optional, Union
-import pickle
-from feathr.definition.anchor import FeatureAnchor
+
 from jinja2 import Template
+
+from feathr.definition.anchor import FeatureAnchor
 from feathr.definition.source import HdfsSource
-import ast
+
 
 # Some metadata that are only needed by Feathr
 FEATHR_PYSPARK_METADATA = 'generated_feathr_pyspark_metadata'
@@ -42,7 +45,7 @@ class _PreprocessingPyudfManager(object):
         # delete the file if it already exists to avoid caching previous results
         for f in [client_udf_repo_path, metadata_path,  pyspark_driver_path]:
             if os.path.exists(f):
-                os.remove(f) 
+                os.remove(f)
 
         for anchor in anchor_list:
             # only support batch source preprocessing for now.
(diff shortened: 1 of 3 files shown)


feathr_project/feathr/udf/_preprocessing_pyudf_manager.py
```diff
@@ -73,23 +76,25 @@ def build_anchor_preprocessing_metadata(anchor_list: List[FeatureAnchor], local_
...
-        code = compile(source=tree, filename='custom.py',mode= 'exec')
-        # https://docs.python.org/3/library/inspect.html see the inspect module for more details
-        # tuple of names other than arguments and function locals. Assume there will be only one function, so will return the first as the name
-        for ele in code.co_consts:
-            # find the first object, that is the str, this will be the name of the function
-            if isinstance(ele, str):
-                return ele
+            raise ValueError("provided code fragment is not a single function")
+
+        return tree.body[0].name
```
xiaoyongzhu, line 97: Thanks! Maybe add a bit explanation in the PR or in the comment on how tree.body[] works?

loomlike: Thank you for the review! tree.body returns the list of `FunctionDef` where if you pass N number of functions in a string, the `len(tree.body) == N`. So tree.body[0] returns the first function definition and `.name` returns its name. (Our expected input fn_str is anyway a single udf, otherwise it throws an exception).

xiaoyongzhu: Sounds good... I mean put it into the comments in the code for easier maintenance for other devs

edwincheung (approved):

jainr (approved):

edwincheung (approved):

PR #734, opened 2022-10-11
Fix synapse errors not print out issue

enya-yx:
## Description

- Currently we get logs from spark pool by requesting 'stdout' folder. It can only get logs when we have 'feathr_pyspark_driver.py' submitted (when applying 'preprocessing'). All error logs should be under 'stderr' folder. So changed the url to get error logs.

## How was this PR tested?

- Tested locally by creating error cases. Got logs like below:

Screen Shot 2022-10-11 at 17 27 14

### DIFF
--- a/feathr_project/feathr/spark_provider/_synapse_submission.py
+++ b/feathr_project/feathr/spark_provider/_synapse_submission.py
@@ -325,7 +325,7 @@ class _SynapseJobRunner(object):
     def get_driver_log(self, job_id) -> str:
         # @see: https://docs.microsoft.com/en-us/azure/synapse-analytics/spark/connect-monitor-azure-synapse-spark-application-level-metrics
         app_id = self.get_spark_batch_job(job_id).app_id
-        url = "%s/sparkhistory/api/v1/sparkpools/%s/livyid/%s/applications/%s/driverlog/stdout/?isDownload=true" % (self._synapse_dev_url, self._spark_pool_name, job_id, app_id)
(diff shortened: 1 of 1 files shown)


feathr_project/feathr/spark_provider/_synapse_submission.py
```diff
@@ -325,7 +325,7 @@ def create_spark_batch_job(self, job_name, main_file, class_name=None,
     def get_driver_log(self, job_id) -> str:
         # @see: https://docs.microsoft.com/en-us/azure/synapse-analytics/spark/connect-monitor-azure-synapse-spark-application-level-metrics
         app_id = self.get_spark_batch_job(job_id).app_id
-        url = "%s/sparkhistory/api/v1/sparkpools/%s/livyid/%s/applications/%s/driverlog/stdout/?isDownload=true" % (self._synapse_dev_url, self._spark_pool_name, job_id, app_id)
+        url = "%s/sparkhistory/api/v1/sparkpools/%s/livyid/%s/applications/%s/driverlog/stderr/?isDownload=true" % (self._synapse_dev_url, self._spark_pool_name, job_id, app_id)
```
Yuqing-cat, line 328: Is it by design to only print stderr without stdout?

enya-yx: We want to get error messages here. And all error messages under stdout are also contained in stderr.

Yuqing-cat: I see, then the function name `get_driver_log` seems not accurate here. Should be `get_driver_log_err()`. But it's ok in this PR.

Yuqing-cat (approved):

xiaoyongzhu (approved):

jainr (approved):
Does this also mean we will get fewer log lines while running notebooks, and will only get errors and nothing for successful jobs?

PR #746, opened 2022-10-13
Unify Python SDK Build Version and decouple Feathr Maven Version

### LINKED ISSUE #745: [FR] Build Version should be configurable in one place
### Willingness to contribute

### Feature Request Proposal

Current Feathr requires version bump in several places. Like in this PR: #722
Especially, the version of Maven Jar should not be hard coded and need to be aligned with Feathr package version. 

### Motivation

To improve build and release process.

### Details

_No response_

- [X] `Computation Engine`: The computation engine that execute the actual feature join and generation work. Mostly in Scala and Spark.

Yuqing-cat:
## Description

Resolves #745 
This PR only changes the build and release experience:
1. use `version.py` to be the single source of truth for Feathr Python SDK version.
2. deprecate origin `feathr.__version__` functions using `pkg_resources`
3. use environment variable `FEATHR_MAVEN_VERSION` to enable different version of MAVEN_ARTIFACT used by Feathr Client.
4. use environment variable `FEATHR_MAVEN_VERSION` in `build.sbt` to make default Maven Version configurable.
5. update `overall_release_guide.md` related components. Broken link Alert is as expected given `version.py` is not checked in to main branch yet. 

## How was this PR tested?

This PR only changes the build and release experience. 


xiaoyongzhu (dismissed):

xiaoyongzhu:
Thanks for the PR. we should also document the environment variable in the docs so folks can refer to it.

Yuqing-cat:
@loomlike, I make some updates to `feathr.__version__`, please help to check if the change make sense to you.

Yuqing-cat:
> Thanks for the PR. we should also document the environment variable in the docs so folks can refer to it.

Indeed! As the config is highly coupled with build and release process, I suggest to expose it to users after a round of release test. Doc Issue https://github.com/feathr-ai/feathr/issues/747 is created to track this.

jainr (commented):
Thanks for the PR, so if I understand correctly currently we have to change build.sbt (for Maven) and setup.py (fpr PyPi) for Java and Python packages respectively.

After this change
Python version would be controlled by version.py file
For Java version user needs to set FEATHR_MAVEN_VERSION on their local machine so it gets passed around to build.sbt?

I am trying to understand the value add here, we still need to update in two places.

Yuqing-cat:
> Thanks for the PR, so if I understand correctly currently we have to change build.sbt (for Maven) and setup.py (fpr PyPi) for Java and Python packages respectively.
> 
> After this change Python version would be controlled by version.py file For Java version user needs to set FEATHR_MAVEN_VERSION on their local machine so it gets passed around to build.sbt?
> 
> I am trying to understand the value add here, we still need to update in two places.

Yes, we still need to update in two places. 
- Python version and the default Maven Package version in Python SDK is control by `version.by`
- Maven Package version in Python SDK can be modified by `FEATHR_MAVEN_JAR`. This is a learning from our last release. User may need to roll back only maven package rather than whole Python package. 
- Java package is controlled by `build.sbt`, which can also be configured by `FEATHR_MAVEN_JAR`.

It's feasible to unify these two settings in a build / release script by set `FEATHR_MAVEN_JAR` = feathr.__version__.

jainr (dismissed):

xiaoyongzhu (approved):

xiaoyongzhu:
LGTM. Please also help fix the conflicts @Yuqing-cat

xiaoyongzhu:
@Yuqing-cat can you also help resolve the conficts?

xiaoyongzhu (approved):

blrchen (approved):

PR #748, opened 2022-10-13
Support SWA with groupBy to 1d tensor conversion

jaymo001:
## Description 
Support converting SWA feature with groupby to 1d dense tensor automatically by keeping the value array and ignore the index array.

This enables the following use case:

  aggregationFeatures: {
    source: idSource
    key: gid
    features: {
      groupedIds: {
        def: "mid"
        groupBy: "mid"
        aggregation: LATEST
        window: 9d
        type: {
            type: TENSOR
            tensorCategory: DENSE
            dimensionType: [INT]
            valType: STRING
        }
      }
    }
  }

## How was this PR tested? 
Add unit test
No

### DIFF
--- a/src/main/scala/com/linkedin/feathr/offline/transformation/FDSConversionUtils.scala
+++ b/src/main/scala/com/linkedin/feathr/offline/transformation/FDSConversionUtils.scala
@@ -253,6 +253,8 @@ private[offline] object FDSConversionUtils {
         values.asScala.toArray
       case values: mutable.WrappedArray[Any] =>
         if (values.nonEmpty && values(0).isInstanceOf[GenericRowWithSchema]) {
+          // Assuming the result is returned by SWA feature with groupBy, hence keeping only the
+          // feature value as an array and dropping the index info.
           values.asInstanceOf[mutable.WrappedArray[GenericRowWithSchema]].map(v => v.get(v.size - 1)).toArray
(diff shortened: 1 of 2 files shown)


src/main/scala/com/linkedin/feathr/offline/transformation/FDSConversionUtils.scala
```diff
@@ -253,7 +252,11 @@ private[offline] object FDSConversionUtils {
       case values: util.ArrayList[Any] =>
         values.asScala.toArray
       case values: mutable.WrappedArray[Any] =>
-        values.asInstanceOf[mutable.WrappedArray[Any]].toArray
+        if (values.nonEmpty && values(0).isInstanceOf[GenericRowWithSchema]) {
```
bozhonghu, line 255: Can we add a comment here documenting this behavior.

bozhonghu (approved):
Synced with Jay offline. Better to do this via adding support for UNION in SWA imo but for now we can add this change in auto-tz here. Minor comment is to document this behavior in code comment clarifying we will drop the indices and concat the values.

bozhonghu (approved):
LGTM

rakeshkashyap123:
LGTM

rakeshkashyap123 (approved):

Issue #716, opened 2022-09-29
[BUG] get_offline_feature ignores `parquet` output file option


loomlike:
### Willingness to contribute

No. I cannot contribute a bug fix at this time.

### Feathr version

0.8.0

### System information

- **OS Platform and Distribution (e.g., Linux Ubuntu 20.0)**: both on Linux Ubuntu 20 and Databricks
- **Python version**: 3.10
- **Spark version, if reporting runtime issue**:

### Describe the problem

get_offline_feature always write into avro regardless of the execution config.

### Code to reproduce bug

Run:
```
get_offline_feature(
    execution_configurations=SparkExecutionConfiguration({
        "spark.feathr.inputFormat": "parquet",
        "spark.feathr.outputFormat": "parquet",
    }),
    ....
)
```
still write file as avro

### What component(s) does this bug affect?

- [ ] `Python Client`: This is the client users use to interact with most of our API. Mostly written in Python.
- [ ] `Computation Engine`: The computation engine that execute the actual feature join and generation work. Mostly in Scala and Spark.
- [ ] `Feature Registry API`: The frontend API layer supports SQL, Purview(Atlas) as storage. The API layer is in Python(FAST API)
- [ ] `Feature Registry Web UI`: The Web UI for feature registry. Written in React

loomlike:
@xiaoyongzhu  Can you reopen this issue? I found this bug exists at Databricks (`_FeathrDatabricksJobLauncher`) too. My previous PR #729 only fixes local spark's.

I'm working on fixing this bug on Databricks. I'll check Synapse too once I get it works on Databricks.

blrchen:
> @xiaoyongzhu Can you reopen this issue? I found this bug exists at Databricks (`_FeathrDatabricksJobLauncher`) too. My previous PR #729 only fixes local spark's.
> 
> I'm working on fixing this bug on Databricks. I'll check Synapse too once I get it works on Databricks.

Hello @loomlike this bug has been reopened

Issue #707, opened 2022-09-27
[BUG] `job_utils.get_result_df` throws an error when using local spark


loomlike:
### Willingness to contribute

Yes. I can contribute a fix for this bug independently.

### Feathr version

0.8.0

### System information

- **OS Platform and Distribution (e.g., Linux Ubuntu 20.0)**: Linux 20.04
- **Python version**: 3.10
- **Spark version, if reporting runtime issue**: 3.2.2

### Describe the problem

local spark launcher `_FeathrDLocalSparkJobLauncher` doesn't implement `download_result` so that the `job_utils.get_result_df`` throws an error when using the local spark.

### What component(s) does this bug affect?

- [X] `Python Client`: This is the client users use to interact with most of our API. Mostly written in Python.
- [ ] `Computation Engine`: The computation engine that execute the actual feature join and generation work. Mostly in Scala and Spark.
- [ ] `Feature Registry API`: The frontend API layer supports SQL, Purview(Atlas) as storage. The API layer is in Python(FAST API)
- [ ] `Feature Registry Web UI`: The Web UI for feature registry. Written in React

PR #756, opened 2022-10-18
Refine example notebooks

### LINKED ISSUE #749: [FR] Refine notebook examples
### Willingness to contribute

### Feature Request Proposal

Refactoring an example notebook to describes basic feathr functionalities with a proper test.

### Motivation

Current example notebooks contain too many unused configurations and are not tested.

### Details

This will cover:
- Clean up and refactor the NYC taxi fare prediction notebook
- Add a sample dataset loader
- Add spark dataframe result output utility 
- Add notebook dependencies to setup.py

loomlike:
## Description

Resolves #749, #716, #707

## How was this PR tested?

Added unit tests and notebook sample test (requires papermill and scrapbook packages -- added to setup.py's extra dependencies of "notebook" and "dev").


feathr_project/feathr/utils/platform.py
```diff
@@ -0,0 +1,38 @@
...
+
+def is_jupyter() -> bool:
+    """Check if the module is running on Jupyter notebook/console.
+
+    Returns:
+        bool: True if the module is running on Jupyter notebook or Jupyter console, False otherwise.
+    """
+    try:
+        shell_name = get_ipython().__class__.__name__
+        if shell_name == "ZMQInteractiveShell":
```
xiaoyongzhu, line 15: Sounds this is hard coded? `ZMQInteractiveShell `

feathr_project/test/samples/test_notebooks.py
```diff
@@ -0,0 +1,56 @@
...
+    .parent     # .../feathr_project
+    .parent     # .../feathr (root of the repo)
+    .joinpath("docs", "samples")
+)
+NOTEBOOK_PATHS = {
+    "nyc_taxi_demo": str(SAMPLES_DIR.joinpath("nyc_taxi_demo.ipynb")),
+}
+
+
+@pytest.mark.notebooks
```
xiaoyongzhu, line 25: After adding this, sounds like it's possible to run tests on the notebooks?

xiaoyongzhu:
Thanks @loomlike ! feathr_project/feathrcli/data/feathr_user_workspace/features/non_agg_features.py for those python files, can we keep them? I think @jaymo001 is using those extensively

feathr_project/feathr/datasets/nyc_taxi.py
```diff
@@ -0,0 +1,91 @@
...
+from urllib.parse import urlparse
+
+import pandas as pd
+from pyspark.sql import DataFrame, SparkSession
+
+from feathr.datasets.utils import maybe_download
+from feathr.utils.platform import is_databricks
+
+
+NYC_TAXI_SMALL_URL = (
```
Yuqing-cat, line 13: Shall we move it to `constants.py` or new a `datasets/constants.py`? It will be useful if we extend to more datasets.

feathr_project/feathr/utils/job_utils.py
```diff
@@ -1,77 +1,187 @@
...
+            Default to `avro` if not specified.
+        res_url: Result URL to download files from. Note that this will not block the job so you need to make sure
+            the job is finished and the result URL contains actual data.
+        local_cache_path (optional): Specify the absolute download path. if the user does not provide this,
+            the function will create a temporary directory.
+
+    Returns:
+        Spark DataFrame
+    """
+    return get_result_df(client, data_format, res_url, local_cache_path, spark=spark)
```
Yuqing-cat, line 61: What's the use case of `get_result_spark_df()` and `get_result_pandas_df()` that `get_result_df()` is not enough?

feathr_project/feathr/utils/job_utils.py
```diff
@@ -1,77 +1,187 @@
...
-        files =  glob.glob(os.path.join(local_dir_path, '*.parquet'))
+        result_df = _read_files_to_pandas_df(
+            dir_path=local_cache_path.replace("dbfs:", "/dbfs"),  # replace to python path if spark path is provided.
+            data_format=data_format,
+        )
+
+    return result_df
+
+
+def _read_files_to_pandas_df(dir_path: str, data_format: str = "avro") -> pd.DataFrame:
```
Yuqing-cat, line 140: Add some constrains to validate "dir_path" will make this function more robust. Given the above 
```
dir_path=local_cache_path.replace("dbfs:", "/dbfs"),  # replace to python path if spark path is provided.
```

feathr_project/feathr/utils/platform.py
```diff
@@ -0,0 +1,38 @@
...
+    try:
+        if str(Path(".").resolve()) == "/databricks/driver":
+            return True
+        else:
+            return False
+    except NameError:
+        return False
+
+
+# TODO maybe add is_synapse()
```
Yuqing-cat, line 38: maybe add is_aml() as well

Yuqing-cat (commented):
Amazing contribution Jun! I love the idea of this refine PR. 
Doesn't have time to verify the deletion in `feathrcli` folder but the rest looks good.

feathr_project/feathr/utils/platform.py
loomlike: This is a bit hacky but works. Ref is: https://stackoverflow.com/questions/15411967/how-can-i-check-if-code-is-executed-in-the-ipython-notebook. I'll add the reference to the function docstring too.

Pre-loaded module `get_ipython()` tells you whether you are running inside ipython or not, and `ZMQInteractiveShell` tells you if this is an interactive mode (notebook).

feathr_project/test/samples/test_notebooks.py
loomlike: With a custom marker, you can control which tests you want to run or not.
E.g. I used a custom marker `notebooks`. To run only the tests marked by this marker,

`pytest -m notebooks` 

Or to exclude,
`pytest -m "not notebooks"`

loomlike:
> feathr_project/feathrcli/data/feathr_user_workspace/features/non_agg_features.py for those python files, can we keep them? I think @jaymo001 is using those extensively

Can you tell me which files & examples use them? Do you think we may move them under *docs/samples* instead?

feathr_project/feathr/utils/job_utils.py
loomlike: `get_result_spark_df` and `get_result_pandas_df` are just an alias to `get_result_df` with more explicit names. I think a function returning two different types of objects depending on arguments are not a good practice and too implicit, but I keep `get_result_df` just because I did not want to remove the previous api which is already in our examples.

feathr_project/feathr/utils/job_utils.py
loomlike: If you're okay, I would leave this function as it is (without extra check of the path) without having an extra check. As the name's prefix suggested (starting with an underscore `_`), this function is intended to be hidden and only used by public functions of this module. The public function who calls this function already checks if the path contains `dbfs`.

feathr_project/feathr/utils/platform.py
loomlike: Do you know if we have a specific use-case and/or modules that only works on AzureML notebook?

Yuqing-cat: I remember there were some conversation with CSE to debug with AML notebook. But I don't think we have modules only works on AML.

feathr_project/feathr/utils/job_utils.py
Yuqing-cat: I see. If you want to expose these two utils functions to endusers, don't forget to add it into `https://github.com/feathr-ai/feathr/blob/main/feathr_project/feathr/__init__.py`.

feathr_project/feathr/utils/platform.py
loomlike: cool. Then let's add is_aml once we have modules just work on AML.

xiaoyongzhu:
@loomlike yeah @jaymo001 is using this in a slightly different way, basically they are trying to store the python files in the "workspace"

feathr_project/feathr/utils/platform.py
xiaoyongzhu: @loomlike interesting to learn that. Can you put that in the comments so others can follow?

loomlike:
Per discussion w/ @Yuqing-cat I'm making changes to `pull_request_push_test.yml` to install extra dependencies via `pip install -e .[all]`. Hope that fixes some of the test errors due to the missing packages.

feathr_project/feathr/utils/job_utils.py
loomlike: I would like to defer this because the functions I added are not the core functions so that it's okay not able to access via the root module `feathr` and also I don't want to make the root module's name space too heavy.

loomlike:
> @loomlike can we also keep this file? feathr_project/feathrcli/data/feathr_user_workspace/feathr_config.yaml
> 
> Also please fix the conflicts.
@xiaoyongzhu done! Thank you for the review!

blrchen (approved):

xiaoyongzhu (approved):

PR #763, opened 2022-10-18
Serialization bug due to version incompatibility between azure-core and msrest

### LINKED ISSUE #761: [BUG] Serialization Error from Azure Core when installing feathr
### Willingness to contribute

### Feathr version

0.8.0

### System information

- **Python version**:
azure-core=1.22.1
msrest==0.7.0

### Describe the problem

It seems it happens due to version mismatch of merest which automatically gets upgraded with azure cor

https://github.com/Azure/azure-sdk-for-python/issues/24765

The root cause is msrest version (0.7.0+) is incompatible with azure-core version (1.22.1). Need to adjust either of them.

### Tracking information

_No response_

### Code to reproduce bug

_No response_

jainr:
## Description
While installing Feathr we pin to specific version of azure-core (1.22.1) due to prior issues on databricks. This version of azure-core only works with msrest (0.6.1) version.
Some environment might have higher version of azure-core pre-installed (Say 1.24.0) which comes with a higher version of 
msrest. As a result we need to pin the version of msrest to 0.6.1 so it works with pinned version of azure-core (1.22.1)

Resolves #761 

## How was this PR tested?
Tested in local environment and made sure right version of msrest gets installed. 
Screen Shot 2022-10-18 at 7 13 10 AM


feathr_project/feathr/udf/_preprocessing_pyudf_manager.py
```diff
@@ -53,6 +53,8 @@ def build_anchor_preprocessing_metadata(anchor_list: List[FeatureAnchor], local_
                 continue
             preprocessing_func = anchor.source.preprocessing
             if preprocessing_func:
+                print("Preprocessing function defined")
```
blrchen, line 56: Should this debug code get removed before merge?

jainr: Good catch, fixed. It was left over from main branch in my fork.

Yuqing-cat (approved):

feathr_project/feathr/udf/_preprocessing_pyudf_manager.py
jainr: Yes, removed.

blrchen (approved):

Uditanshu1612:
Hi,
Is the issue is resolved? I'm also facing the same issue while importing a package.
Currently My azure-core is a version of 1.26.0 so while executing the module using this version of azure-core it throws me an error

image


So here by observing the error I get to know that module requires an azure-core of version <=1.22.1. So, if downgrade the version to 1.22.1 or less than that, then I'm facing an error

image


did anyone have the solution how to resolve the error and successfully import the module?

Uditanshu1612:
Hi,
Is the issue is resolved? I'm also facing the same issue while importing a package.
Currently My azure-core is a version of 1.26.0 and msrest is a version of 0.6.21 so while executing the module using this version of azure-core it throws me an error

image


So here by observing the error I get to know that module requires an azure-core of version <=1.22.1. So, if downgrade the version to 1.22.1 or less than that, with mrest = 0.6.1 then I'm facing an error

image


did anyone have the solution how to resolve the error and successfully import the module?

jainr:
I

> Hi, Is the issue is resolved? I'm also facing the same issue while importing a package. Currently My azure-core is a version of 1.26.0 and msrest is a version of 0.6.21 so while executing the module using this version of azure-core it throws me an error
> 
> image
> 
> So here by observing the error I get to know that module requires an azure-core of version <=1.22.1. So, if downgrade the version to 1.22.1 or less than that, with mrest = 0.6.1 then I'm facing an error
> 
> image
> 
> did anyone have the solution how to resolve the error and successfully import the module?

It seems this is a problem with version mismatch between azure-core and azure-cli. See https://github.com/Azure/azure-cli/issues/23267,  I also see you posted it there too.

What version of azure-cli are you using, could you try to downgrade it to <=2.37.0 and see if the issue still persisits?

PR #769, opened 2022-10-20
Model

hyingyang-linkedin:
## Description
This RB is the following RB of [Add Data Models](https://github.com/feathr-ai/feathr/pull/659)
Based on [doc](https://microsoft-my.sharepoint.com/:w:/g/personal/hayang_linkedin_biz/ETJAXn8vo2ZPhyeE4H01wcABsDjmcCDfM6xgFPOCDPFz5w?e=dGKR73)

Changes in this RB:
## How was this PR tested?
N/A

- [x ] No. You can skip the rest of this section.


registry/data-models/common/models.py
```diff
@@ -0,0 +1,139 @@
...
+    Base model for all functions
+    """
+    expression: str  # Expression in str format
+
+
+class MvelExpression(Function):
+    """
+    An expression in MVEL language.
+    """
+    mvel: str  # The MVEL expression
```
LeBenHL, line 103: This would mean that this model has both a `mvel` and `expression` field right? Is this intended?

registry/data-models/common/models.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+class UserDefinedFunction(Function):
```
LeBenHL, line 106: Curious, how would `expression` be filled in this case?

registry/data-models/models.py
```diff
@@ -0,0 +1,152 @@
+from registry.data-models.transformation.models import *
+from registry.data-models.common.models import SemanticVersion, FeathrModel, Function
+from typing import Optional
+from pydantic import BaseModel
+
+
+"""
```
LeBenHL, line 7: We created this file before right in the previous PR? How come this is showing us as a new file rather than updates on the old one?

registry/data-models/models.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+    id: str  # id of a project
+
+
+class Source(FeathrModel):
+    """
+    Source of the feature.
+    It defines where the feature is extracted or derived from.
+    """
+    sourceRef: Optional[str]  # Represents the identifier of a Source object.
```
LeBenHL, line 52: Is this really optional?

registry/data-models/models.py
```diff
@@ -0,0 +1,152 @@
...
+    alias: Optional[str]  # A feature's alias to be used in transformation function.
+
+
+class MultiFeatureSource(Source):
+    """
+    Feature sources of the feature.
+    It defines one to many features where the feature is derived from.
+    """
+    sources: List[FeatureSource]  # All source features which the feature is derived from
+    pass
```
LeBenHL, line 79: Remove this pass

registry/data-models/transformation/models.py
```diff
@@ -0,0 +1,84 @@
...
+    size: int  # Represents the duration of the window.
+    unit: WindowTimeUnit
+
+
+class SlidingWindowAggregation(Function):
+    """
+    Sliding window aggregation produces feature data by aggregating a collection of data within a given time
+    interval into an aggregate value. It ensures point-in-time correctness, when joining with label data,
+    it looks back the configurable time window from each entry's timestamp and compute the aggregate value.
+    This class can be extended to support LateralView in aggregation.
```
LeBenHL, line 50: @xiaoyongzhu. Scala code supports lateral view, even though current Python API for Frame Feature Definition does not. 

Do you want to still skip the lateral view in the Feature Registry data model too and add it one once the functionality is exposed in the Feature Definition API?

registry/data-models/common/models.py
```diff
@@ -0,0 +1,139 @@
...
+
+class FeathrModel(BaseModel):
+    """
+    Base model for feathr entity which will be displayed in Feathr UI
+    """
+    displayName: str  # name of the entity showed on UI
+    typeName: str  # type of entity in str format, will be displayed in UI
+    # Represents zero, one or multiple keyPlaceholderRefs which are used as the
+    # identifiers to reference KeyPlaceholders of the FeatureSource
+    keyPlaceholderRefs: List[str] = []
```
LeBenHL, line 139: Do we need this as a common field? Feels like this should be a more specific field to some entity

registry/data-models/transformation/models.py
```diff
@@ -0,0 +1,84 @@
...
+    representation and reduce its dimensionality.
+    """
+    MAX_POOLING = "max_pooling"  # Max pooling is done by applying a max filter to (usually) non-overlapping subregions of the initial representation
+    MIN_POOLING = "min_pooling"  # Min pooling is done by applying a min filter to (usually) non-overlapping subregions of the initial representation.
+    AVG_POOLING = "avg_pooling"  # Average pooling is done by applying a average filter to (usually) non-overlapping subregions of the initial representation
+
+
+class WindowTimeUnit(Enum):
+    """
+     Represents a unit of time.
```
LeBenHL, line 29: nit: fix spacing

LeBenHL (changes requested):

registry/data-models/common/models.py
hyingyang-linkedin: Internally we can use  mvel as experssion.

registry/data-models/common/models.py
hyingyang-linkedin: Since there is clazz name and parameters of UserDefinedFunction, we might need to add these two fields in experssion. I didn't implement this part as I am not sure how we want to show this in UI

registry/data-models/models.py
hyingyang-linkedin: I had some conflicts earlier. I just resolved all the conflicts, diff should be right now.

registry/data-models/transformation/models.py
hyingyang-linkedin: Thanks. fixed.

registry/data-models/models.py
hyingyang-linkedin: Thanks for catching it.

registry/data-models/common/models.py
LeBenHL: I think I understand now. `mvel` will be used internally for execution but `expression` used only for display purposes.

Is this correct?

donegjookim: @LeBenHL - I think so.
@hyingyang-linkedin - If the intention is to identify the type of expression, what do you think about having `type` attribute on `Function` class, similar to the current Feathr's `Entity` class

registry/data-models/common/models.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    clazz: Clazz  # Reference to the class that implements the user defined function.
+    parameters: Dict[str, json] = {}  # This field defines the custom parameters of the user defined function
+
+
+class SparkSqlExpression(Function):
+    """
+    An expression in Spark SQL.
+    """
+    sql: str  # Spark SQl expression
```
donegjookim, line 118: I think my point of reusing `expression` and adding in `type` attribute applies here as well

donegjookim: The benefit of doing that is that we can dynamically render common attributes in the UI for newly added classes without much (if at all) code change

LeBenHL (approved):

windoze (approved):
Looked into the updated model classes, they look good to me.

PR #776, opened 2022-10-25
Fix Windows compatibility issues

xiaoyongzhu:
## Description

This PR solves a few compatibility issues when using Windows as the client development env.

Specifically, when Feathr tries to upload the resources to cloud environment, the cloud environment expects all the paths to be UNIX style; however if the client is Windows, we sometimes use `os.path.join` which will yield paths like this:

`dbfs:/feathrazure_cijob_snowflake_9_30_157692\auto_generated_derived_features.conf`
which has mixed UNIX/Windows style paths, and this will not be recognized by the system on the cloud.

This PR also adds docs for a few missing configurations (https://github.com/feathr-ai/feathr/issues/775)

- [X] No. You can skip the rest of this section.


feathr_project/feathr/spark_provider/_databricks_submission.py
```diff
@@ -69,23 +69,25 @@ def upload_or_get_cloud_path(self, local_path_or_http_path: str):
         """
         src_parse_result = urlparse(local_path_or_http_path)
         file_name = os.path.basename(local_path_or_http_path)
-        # returned paths for the uploaded file
-        returned_path = os.path.join(self.databricks_work_dir, file_name)
+        # returned paths for the uploaded file. Note that we cannot use os.path.join here, since in Windows system it will yield paths like this:
+        # dbfs:/feathrazure_cijob_snowflake_9_30_157692\auto_generated_derived_features.conf, where the path sep is mixed, and won't be able to be parsed by databricks.
+        # so we force the path to be Linux style here.
+        cloud_dest_path = self.databricks_work_dir + "/" + file_name
```
Yuqing-cat, line 75: Line 72-75 repeated in line 113-116. Would be better if we would merge them.

Yuqing-cat (approved):

feathr_project/feathr/spark_provider/_databricks_submission.py
xiaoyongzhu: Good point and I also tried to do that, but the key issue is that in line 101-102, we need to upload multiple files (in a folder). So either we duplicate the code in line 101-102, or duplicate in 113-116...

enya-yx (approved):
LGTM

Yuqing-cat (approved):

PR #808, opened 2022-11-01
Fix avro files read failure under timePartitionPattern paths 

enya-yx:
## Description

- Modify test cases to test 'timePartitionPattern' e2e


Yuqing-cat (approved):

feathr_project/test/test_azure_spark_e2e.py
```diff
@@ -393,6 +393,66 @@ def test_feathr_materialize_to_aerospike():
...
+                backfill_time=backfill_time)
+    client_producer.materialize_features(settings)
+    # assuming the job can successfully run; otherwise it will throw exception
+    client_producer.wait_job_to_finish(timeout_sec=Constants.SPARK_JOB_TIMEOUT_SECONDS)
+
+    # download result and just assert the returned result is not empty
+    # by default, it will write to a folder appended with date
+    res_df = get_result_df(client_producer, "avro", output_path + "/df0/daily/2020/05/20")
+    assert res_df.shape[0] > 0
+
```
xiaoyongzhu, line 429: Can we add a bit comment here on how is this test being setup? Looks like what we are doing is try to generate a bunch of files in a certain folder with some path patterns, and read those files?

enya-yx: Correct. Since we need to make sure there're some data sources under the paths we want to test (eg. basePath + {yyyy/mm/dd}), so create these data sources in advance by running a materialize job for both synapse and databricks.

enya-yx:
> Can docs get updated to give samples on how to use this new params?

The comments under definition of 'HdfsSource' is the related document update and it can be shown through API document link we exposed.

blrchen:
> > Can docs get updated to give samples on how to use this new params?
> 
> The comments under definition of 'HdfsSource' is the related document update and it can be shown through API document link we exposed.

Yes, that is what i meant, we should put that in docs because user normally do not read our test code, they read docs.

enya-yx:
> > > Can docs get updated to give samples on how to use this new params?
> > 
> > 
> > The comments under definition of 'HdfsSource' is the related document update and it can be shown through API document link we exposed.
> 
> Yes, that is what i meant, we should put that in docs because user normally do not read our test code, they read docs.

The comment was already added in the file 'feathr_project/feathr/definition/source.py'

blrchen:
> > > > Can docs get updated to give samples on how to use this new params?
> > > 
> > > 
> > > The comments under definition of 'HdfsSource' is the related document update and it can be shown through API document link we exposed.
> > 
> > 
> > Yes, that is what i meant, we should put that in docs because user normally do not read our test code, they read docs.
> 
> The comment was already added in the file 'feathr_project/feathr/definition/source.py'

I see why you are confused here, might be I didn't explain clearly enough, I meant public docs like this one https://feathr-ai.github.io/feathr/how-to-guides/feathr-input-format.html should be updated for use directories as input source.

blrchen (approved):

xiaoyongzhu (approved):

PR #838, opened 2022-11-07
Decouple build feature code

xiaoyongzhu:
## Description

This PR decouples the `save_to_feature_config_from_context` method from the registry code, given:
1. It doesn't make sense to have this method in the registry code
2. This method introduces unnecessary dependencies for registry, which makes a few other functionalities have dependencies on feature registry, such as local spark runner/multi cloud support.


feathr_project/feathr/client.py
```diff
@@ -169,8 +176,23 @@ def __init__(self, config_path:str = "./feathr_config.yaml", local_workspace_dir
...
+        azure_purview_name = self.envutils.get_environment_variable_with_default('feature_registry', 'purview', 'purview_name')
+        if registry_endpoint:
+            self.registry = _FeatureRegistry(self.project_name, endpoint=registry_endpoint, project_tags=project_registry_tag, credential=credential)
+        elif azure_purview_name:
+            registry_delimiter = self.envutils.get_environment_variable_with_default('feature_registry', 'purview', 'delimiter')
+            # initialize the registry no matter whether we set purview name or not, given some of the methods are used there.
+            self.registry = _PurviewRegistry(self.project_name, azure_purview_name, registry_delimiter, project_registry_tag, config_path = config_path, credential=credential)
+        else:
+            # no registry configured
+            logger.info("Feathr registry is not configured. Consider setting the Feathr registry component for a richer feature store experience.")
```
xiaoyongzhu, line 194: for richer feature store experience

feathr_project/feathr/registry/_feathr_registry_client.py
```diff
@@ -196,185 +196,6 @@ def _post(self, path: str, body: dict) -> dict:
     def _get_auth_header(self) -> dict:
         return {"Authorization": f'Bearer {self.credential.get_token("https://management.azure.com/.default").token}'}
```
aabbasi-hbo, line 197: @xiaoyongzhu I believe there is another azure coupling here as well

xiaoyongzhu: Good point, let me update

feathr_project/feathr/client.py
```diff
@@ -169,8 +176,23 @@ def __init__(self, config_path:str = "./feathr_config.yaml", local_workspace_dir
 
         self.secret_names = []
 
+        # initialize config helper
+        self.config_helper = FeathrConfigHelper()
+
         # initialize registry
-        self.registry = default_registry_client(self.project_name, config_path=config_path, credential=self.credential)
+        self.registry = None
```
Yuqing-cat, line 183: Shall we still wrap line [183,194] into a registry initialization function?

feathr_project/feathr/registry/_feathr_registry_client.py
xiaoyongzhu: @aabbasi-hbo I was looking at it and feel it might make sense to have a follow up PR. Basically I think the current auth is coupled with Azure credentials, so in that PR I think I'll separate the auth part.

I have opened this github issue to track this:
https://github.com/feathr-ai/feathr/issues/873

feathr_project/feathr/client.py
xiaoyongzhu: > Shall we still wrap line [183,194] into a registry initialization function?

Good point, I tried to do that as well but it needs to pass around quite a few parameters like tags etc., so eventually I think making it part of the main body might be better

feathr_project/feathr/definition/config_helper.py
```diff
@@ -0,0 +1,193 @@
...
+from feathr.registry.feature_registry import FeathrRegistry
+from feathr.definition.repo_definitions import RepoDefinitions
+from pathlib import Path
+from jinja2 import Template
+import sys
+from feathr.utils._file_utils import write_to_file
+import importlib
+import os
+
+class FeathrConfigHelper(object):
```
Yuqing-cat, line 22: Shall we add some unit test for this large config helper class?

Yuqing-cat (commented):
Looks good to me, add some comments.

feathr_project/feathr/client.py
```diff
@@ -170,10 +174,23 @@ def __init__(self, config_path:str = "./feathr_config.yaml", local_workspace_dir
 
         self.secret_names = []
 
-        # initialize registry
-        self.registry = default_registry_client(self.project_name, config_path=config_path, credential=self.credential)
+        # initialize config helper
+        self.config_helper = FeathrConfigHelper()
 
-        logger.info(f"Feathr Client {get_version()} initialized successfully")
```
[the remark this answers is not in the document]
xiaoyongzhu: Good point, I think it was caused by a code merge issue so I have to manually recover a few lines, and I guess I missed this part.

feathr_project/feathr/definition/config_helper.py
xiaoyongzhu: Yeah the focus of this PR is to restructure the code, the min bar is the E2E test should work, and we can add unit test in the follow up PR

PR #852, opened 2022-11-12
Expose deletion API for projects/features

### LINKED ISSUE #842: [FR] Expose the deletion API for features/projects
### Willingness to contribute

### Feature Request Proposal

Expose the deletion API for features/projects

### Motivation

Expose the deletion API for features/projects

### Details

_No response_

aabbasi-hbo:
## Description

Resolves #842 

Exposes the deletion API for features/projects

## How was this PR tested?

Updated unit tests


aabbasi-hbo (commented):
@xiaoyongzhu is there a reason why the registry purview client (https://github.com/feathr-ai/feathr/blob/main/feathr_project/feathr/registry/_feature_registry_purview.py) does not call the API specified in (https://github.com/feathr-ai/feathr/blob/main/registry/purview-registry/main.py)? It seems that API functionality has been directly implemented in the client

aabbasi-hbo:
> @xiaoyongzhu is there a reason why the registry purview client (https://github.com/feathr-ai/feathr/blob/main/feathr_project/feathr/registry/_feature_registry_purview.py) does not call the API specified in (https://github.com/feathr-ai/feathr/blob/main/registry/purview-registry/main.py)? It seems that API functionality has been directly implemented in the client

Nevermind I see that it is deprecated

xiaoyongzhu:
Thanks @aabbasi-hbo for the PR! This is very helpful. 

@windoze @YihuiGuo can you take a look as well?

windoze:
Cascading deletion seems to be a too-dangerous operation, could we have some safer alternatives?
My idea is to add another method to retrieve all downstream, so we can know the correct order to delete entity one by one

aabbasi-hbo:
> 



> Cascading deletion seems to be a too-dangerous operation, could we have some safer alternatives? My idea is to add another method to retrieve all downstream, so we can know the correct order to delete entity one by one

@windoze which cascading operation are you referring to? The deletion uses the BFS function to get all downstream entities and deletes them through another BFS

windoze:
'Cascading" means one delete request could actually delete a lot of entities without any prior notice, which could be too dangerous.
I prefer to add 2 APIs, one to get a list of entities that depend on the specific one in the correct order, and another one to delete one entity at a time, refusing to delete an entity that has any downstream.

aabbasi-hbo:
> 'Cascading" means one delete request could actually delete a lot of entities without any prior notice, which could be too dangerous. I prefer to add 2 APIs, one to get a list of entities that depend on the specific one in the correct order, and another one to delete one entity at a time, refusing to delete an entity that has any downstream.

got it. yeah that makes sense - will update. Thanks for the feedback!

aabbasi-hbo (commented):
@windoze mind reviewing again when you get a chance? thanks!

registry/purview-registry/main.py
```diff
@@ -56,6 +56,22 @@ def get_projects_ids() -> dict:
...
+    downstream_entities = registry.get_dependent_entities(entity_id)
+    return list([e.to_dict() for e in downstream_entities])
+
+@router.delete("/entity/{entity}")
+def delete_entity(entity: str):
+    entity_id = registry.get_entity_id(entity)
+    downstream_entities = registry.get_dependent_entities(entity_id)
+    if len(downstream_entities) > 0:
+        raise HTTPException(
+           status_code=500, detail=f"""Entity cannot be deleted as it has downstream/dependent entities.
```
windoze, line 71: Better to use a status code other than 500, my recommendation is "412 precondition failed".

registry/sql-registry/main.py
```diff
@@ -86,6 +86,22 @@ def get_projects_ids() -> dict:
...
+    downstream_entities = registry.get_dependent_entities(entity_id)
+    return list([e.to_dict() for e in downstream_entities])
+
+@router.delete("/entity/{entity}")
+def delete_entity(entity: str):
+    entity_id = registry.get_entity_id(entity)
+    downstream_entities = registry.get_dependent_entities(entity_id)
+    if len(downstream_entities) > 0:
+        raise HTTPException(
+            status_code=500, detail=f"""Entity cannot be deleted as it has downstream/dependent entities.
```
windoze, line 101: Better to use a status code other than 500, my recommendation is "412 precondition failed".

windoze (commented):
It looks good to me, thanks for a great job.

windoze:
> @windoze mind reviewing again when you get a chance? thanks!

Sorry for the delay, I was working on something else.
I've put my comments, please take a look.

Thanks for the contribution!

registry/purview-registry/main.py
aabbasi-hbo: updated

registry/sql-registry/main.py
aabbasi-hbo: updated

xiaoyongzhu (approved):
This is a really nice PR, thanks for the contribution!

windoze (approved):

PR #875, opened 2022-11-18
A new compute model with gradle build

rakeshkashyap123:
In this PR:-

We add a new compute model and engine to Feathr for faster performance:-
Compute Model
a. Convert the hocon config model into a data model which is simple to understand, this is stored in feathr-config module.
The hocon config gets converted to
i) AnchorConfig
ii) DerivationConfig
iii) SequentialJoin Config
All the above code is in feathr-config module.
b. The above data models are then converted into a compute graph. A compute graph consists of the following nodes:-
a. DataSource node - An entity to encapsulate the datasource information of a feature.
i) Context - Represents the training dataset information
ii) Table - A table data source contains both a snapshot view and an update log.
iii) Event - An event node contains append-only event logs whose records need to be grouped and aggregated (e.g. counted, averaged, top-K’d) over a limited window of time.
b. Transformation node - An entity to encapsulate all the transformations which are to be done on top of a datasource/other nodes.
c. Lookup node - Maps to the present day sequential join object.
d. Aggregation node - Maps to the currentt SWA join object.
e. External node - A node used as a placeholder while computing other nodes, not to be used by the compute engine.
c. We generate a raw graph using only the feature def config.
d. Once the join config is passed in, we optimize this graph by performing generic graph operations like merge, delete, prune, etc.

All of the above information is in the feathr-compute module.

Compute Engine
a. We pass the above the compute graph into the compute engine which sorts the graph topologically.
b. We group a few related nodes like all SWA nodes, similar transformation function nodes together for faster execution by the engine.
c. Then, we execute the nodes and add the computed features onto the original observation data.

All of the above engine is in the feathr-impl module.
We moved to gradle based build from SBT.
We have made use of PDL data object to store all the data models. This requires us to migrate from SBT to Gradle version.
How was this PR tested?
Currently testing it with multiple flows. The flows either perform faster or have a similar performance. All integration and unit tests pass.

New integration and unit tests have been added.

Does this PR introduce any user-facing changes?
It is completely backward compatible. Also, this would not be the default engine for now, we would continue testing till we achieve complete confidence.


feathr_gradle/feathr-config/build.gradle
```diff
@@ -0,0 +1,88 @@
...
+
+test {
+  maxParallelForks = 1
+  forkEvery = 1
+  // need to keep a lower heap size (TOOLS-296596)
+  minHeapSize = "512m"
+  useTestNG()
+}
+
+java {
```
bozhonghu, line 32: Do we still need to create a jar and publish for every single module here? Initially I wasn't able to create a single build directory and jar but I believe you were able to resolve that. In that case we should remove the publishing code in the gradle files in the individual modules and just have one for the overall project.

PR #877, opened 2022-11-23
add KeyError for key and feature type

### LINKED ISSUE #766: [EN] Add warnings for key types and feature types (since they are different)
### Willingness to contribute

### Feathr version

0.8

### System information

- **Python version**:

### Describe the problem

Currently the Feathr key type and the feature type is different and users usually mix those (we've seen it multiple times). So we should:

1. Add type checks for the type of key (ValueType.INT32) and feature types (from feathr import INT32)
2. give warnings out if users mix those

### Tracking information

_No response_

### Code to reproduce bug

_No response_

Yuqing-cat:
Signed-off-by: Yuqing Wei <weiyuqing021@outlook.com>

## Description

Resolves #766 

## How was this PR tested?

Test `test_dtype.py` added under unit folder with 2 functions:
- test_key_type
- test_feature_type

image

- [x] Yes. Make sure to clarify your proposed changes.

User will get KeyError if the type of Key or Feature is not correct.

### DIFF
--- a/feathr_project/feathr/definition/feature.py
+++ b/feathr_project/feathr/definition/feature.py
@@ -30,6 +30,11 @@ class FeatureBase(HoconConvertible):
                  registry_tags: Optional[Dict[str, str]] = None,
                  ):
         FeatureBase.validate_feature_name(name)
+
+        # Validate the feature type
+        if not isinstance(feature_type, FeatureType):
+            raise KeyError(f'Feature type must be a FeatureType class, like INT32, but got {feature_type}')
+
         self.name = name
         self.feature_type = feature_type
         self.registry_tags=registry_tags
--- a/feathr_project/feathr/definition/typed_key.py
+++ b/feathr_project/feathr/definition/typed_key.py
@@ -20,6 +20,10 @@ class TypedKey:
                  full_name: Optional[str] = None,
                  description: Optional[str] = None,
                  key_column_alias: Optional[str] = None) -> None:
+        # Validate the key_column type
+        if not isinstance(key_column_type, ValueType):
+            raise KeyError(f'key_column_type must be a ValueType, like Value.INT32, but got {key_column_type}')
+        
         self.key_column = key_column
         self.key_column_type = key_column_type
         self.full_name = full_name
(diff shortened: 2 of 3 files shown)


xiaoyongzhu (approved):

xiaoyongzhu:
The test failure should be unrelated IMO

feathr_project/feathr/definition/feature.py
```diff
@@ -30,6 +30,11 @@ def __init__(self,
                  registry_tags: Optional[Dict[str, str]] = None,
                  ):
         FeatureBase.validate_feature_name(name)
+
+        # Validate the feature type
+        if not isinstance(feature_type, FeatureType):
+            raise KeyError(f'Feature type must be a FeatureType class, like INT32, but got {feature_type}')
```
blrchen, line 36: Suggest change to {feature_type} is not a valid feature type, please use one of following valid values: blabla (dump all enum values)

Yuqing-cat: It's feasible for `KeyType`, which is a `enum`, but cannot be applied to `FeatureType` as it's a base class. 

From my perspective, message looks too long to dump all enum values: 

KeyType: (all of them needs to add `ValueType` as prefix)
    UNSPECIFIED = 0
    BOOL = 1
    INT32 = 2
    INT64 = 3
    FLOAT = 4
    DOUBLE = 5
    STRING = 6
    BYTES = 7

FeatureType:
BOOLEAN = BooleanFeatureType()
INT32 = Int32FeatureType()
INT64 = Int64FeatureType()
FLOAT = FloatFeatureType()
DOUBLE = DoubleFeatureType()
STRING = StringFeatureType()
BYTES = BytesFeatureType()
FLOAT_VECTOR = FloatVectorFeatureType()
INT32_VECTOR = Int32VectorFeatureType()
INT64_VECTOR = Int64VectorFeatureType()
DOUBLE_VECTOR = DoubleVectorFeatureType()

windoze: Maybe we should use different sets of names for different types, but that'll be a breaking change.

xiaoyongzhu: > Maybe we should use different sets of names for different types, but that'll be a breaking change.

Agree, that might be ideal but will introduce breaking change. I think we can simply using this check to help users know what's wrong there

PR #894, opened 2022-12-01
Fix broken CI tests and test release code

rakeshkashyap123:
1. Test release branch
2. Exclude avroutil package from feathr
3. Fix broken tests


xiaoyongzhu:
Let's get this in, seems we only have one failing test case and let's fix them in the next PR.

gradle.properties
```diff
@@ -1,3 +1,3 @@
-version=0.11.1-rc1
+version=v0.11.1-rc1
```
blrchen, line 1: Please use v0.10.3-rc5 as next publis release is v0.10.

blrchen:
Thanks @rakeshkashyap123 for fixing CI issues. Suggest update PR title to mention what this PR is for.

I think one of remaining works is update [publish-to-maven.yml](https://github.com/feathr-ai/feathr/blob/main/.github/workflows/publish-to-maven.yml) to publish 4 newly added jar files.

gradle.properties
rakeshkashyap123: Updated.

blrchen (dismissed):

blrchen (approved):

xiaoyongzhu (approved):

PR #906, opened 2022-12-08
Enya/cov ignore registry

enya-yx:
## Description

- Add 'registry_utils' as one of files should be ignored in test coverage report since it's only used by '_feature_registry_purview' which is deprecated.


enya-yx:
Need to solve conflicts with main first

PR #917, opened 2022-12-12
Add FIRST aggregation for look up feature

jaymo001:
## Description
Currently, users are relying on UNION aggregation based look up feature, followed by a derived feature to pick the first value, when there is only one looked up feature value and the user wants to keep the look up feature value type the same as input.
This PR adds FIRST aggregation for look up feature so that users can skip the derived feature definition for such tasks.

## How was this PR tested?
Add test case


bozhonghu (approved):
LGTM. Should make it clear in documentation this should only be used if there is a single value and also make it clear this is only available in offline and not in online.

xiaoyongzhu (approved):

PR #923, opened 2022-12-14
Add use_env_var flag to client

### LINKED ISSUE #922: [FR] Add an option to use config file over env vars.
### Willingness to contribute

### Feature Request Proposal

Give options to read config variables either from env or yaml file.

### Motivation

Currently, feathr client always use env vars regardless of the configured variables in the feathr yaml file.
This behavior is sometimes confusing since the yaml config is explicitly written while the env vars are hidden and implicit.

I think some of customers already get used to use env vars over the config file and thus changing the priority of reading config variables may cause problems to the existing workflows and pipelines.

So I'm proposing introducing a flag to FeathrClient to indicate which variables (env or config yaml file) the feathr client should read.

### Details

_No response_

loomlike:

## Description
Introduce an option to select between env vars and config yaml file.
Feathr client to use explicitly configured yaml file over environment variable if `use_env_var` flag is set to False.

The changes are added as the last argument of the existing functions and set default to True (use env variables) so that
existing codes don't break.

Resolves #922 

## How was this PR tested?

Unit test

- [x] Yes. Make sure to clarify your proposed changes.


feathr_project/feathr/utils/_envvariableutil.py
```diff
@@ -5,87 +5,116 @@
...
         """Gets the environment variable for the variable key.
+        If use_env_vars set to False, it will return the default value from the config file.
+
         Args:
             *args: list of keys in feathr_config.yaml file
-        Return:
-            A environment variable for the variable key. It will retrieve the value of the environment variables in the following order:
+
+        Returns:
+            An environment variable for the variable key. It will retrieve the value of the environment variables in the following order:
```
enya-yx, line 34: Should we change the 'Returns' comments here? I guess if 'use_env_vars' is False, we will not return the environment variable when the key is set.

feathr_project/feathr/utils/_envvariableutil.py
```diff
@@ -5,87 +5,116 @@
...
-        # make it work for lower case and upper case.
-        env_variable = os.environ.get(
-            env_keyword, os.environ.get(upper_env_keyword))
+        """
+        variable_key = "__".join(args)
 
-        # If the key is set in the environment variable, Feathr will use the value of that environment variable
-        if env_variable:
-            return env_variable
+        env_var = self._get_variable_from_env(variable_key)
```
enya-yx, line 42: Compared with checking 'use_env_vars' in the '_get_variable_from_env', I would suggest to check it here, which means if 'use_env_vars' is False, we will not need to call the '_get_variable_from_env'

feathr_project/feathr/utils/_envvariableutil.py
loomlike: yes. I'll update the docstring accordingly. Thank you!

feathr_project/feathr/utils/_envvariableutil.py
loomlike: sounds good!

enya-yx (approved):

feathr_project/feathr/utils/_envvariableutil.py
```diff
@@ -5,39 +5,101 @@
 from azure.core.exceptions import ResourceNotFoundError
 
 class _EnvVaraibleUtil(object):
```
Yuqing-cat, line 7: Find a long living typo: `_EnvVaraibleUtil ` -> `_EnvVariableUtil`

Yuqing-cat (commented):
The enhancement looks reasonable. It will be better if we could make _EnvVariableUtil less complex.

feathr_project/feathr/utils/_envvariableutil.py
```diff
@@ -5,39 +5,101 @@
...
+            use_env_vars (optional): Whether to use os environment variables instead of config file. Defaults to True.
+        """
         self.config_path = config_path
-        # Set to none first to avoid invalid reference
-        self.akv_name = None
-        self.akv_name = self.get_environment_variable_with_default( 'secrets', 'azure_key_vault', 'name')
+        self.use_env_vars = use_env_vars
+
+        self.akv_name = (
+            self._get_variable_from_env("secrets__azure_key_vault__name") or
```
blrchen, line 24: Should `self._get_variable_from_env("secrets__azure_key_vault__name")` respect use_env_vars flag?

feathr_project/feathr/utils/_envvariableutil.py
```diff
@@ -5,39 +5,101 @@
...
+        variable_key = "__".join(args)
+
+        env_var = (
+            (self._get_variable_from_env(variable_key) if self.use_env_vars else None) or
+            self._get_variable_from_file(*args) or
+            (self._get_variable_from_akv(variable_key) if self.akv_name else None)
+        )
+
+        if env_var is None:
+            logger.warning(f"Environment variable {variable_key} doesn't exist in environment variable, YAML config file, and key vault service.")
```
blrchen, line 53: This output might needs to be updated to avoid confusion

blrchen:
Hello @loomlike seems this doc should get updated as well https://feathr-ai.github.io/feathr/how-to-guides/feathr-configuration-and-env.html#configuration-and-environment-variables-in-feathr

Can I ask is there any use case for introducing this new env? Have this new flag can make configuration more flexible but it introduces complexity on configuration code as well.

feathr_project/feathr/utils/_envvariableutil.py
loomlike: good catch!

feathr_project/feathr/utils/_envvariableutil.py
loomlike: make sense. will change that.

loomlike:
> Hello @loomlike seems this doc should get updated as well https://feathr-ai.github.io/feathr/how-to-guides/feathr-configuration-and-env.html#configuration-and-environment-variables-in-feathr
> 
> Can I ask is there any use case for introducing this new env? Have this new flag can make configuration more flexible but it introduces complexity on configuration code as well.

The default behavior is to override the config file values (`use_env_vars=True`) so I think we can keep the document as it is.
Usecases are mostly for us to test with different configuration of the client, since if we change the env var of the test machine for each test case, it's not safe (someone can forget to put the original env var back at the end of the test) as well as parallel testing will not work as they'll share the env vars.

feathr_project/feathr/utils/_envvariableutil.py
```diff
@@ -5,39 +5,101 @@
 from azure.core.exceptions import ResourceNotFoundError
 
 class _EnvVaraibleUtil(object):
-    def __init__(self, config_path):
+    """A utility class to read config variables.
+    If use_env_vars set to False, `get_environment_variable_with_default` will not use os environment variables.
```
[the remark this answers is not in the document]
loomlike: I've change to have `get` and `get_from_env_or_akv` methods. Also changed the function to preload yaml instead of reading the file for each get operation. Let me know if that looks good.

feathr_project/feathr/utils/_envvariableutil.py
blrchen: Sounds good

Yuqing-cat (commented):
Looks nice to me! Thanks for the changes. 
Please also change the file names `envvariableutil.py` & `test_envvariableutil.py` to fit the new class name.

blrchen (approved):

jainr (approved):
Thanks for making the changes and addressing the comments.

PR #929, opened 2022-12-17
Add is_synapse()

loomlike:
## Description

Some notebooks require to check if the notebook is running on synapse. This is a bit hacky way to check the platform but working solution.
More explicit way should be using a spark context, but then we have to pass the spark object to the function as an argument.

## How was this PR tested?

Unit tests for negative cases.
Manual tests (on databricks and synapse) for positive cases.

### DIFF
--- a/feathr_project/feathr/utils/platform.py
+++ b/feathr_project/feathr/utils/platform.py
@@ -1,8 +1,4 @@
-"""Platform utilities.
-Refs: https://github.com/microsoft/recommenders/blob/main/recommenders/utils/notebook_utils.py
-"""
 import os
-from pathlib import Path
 
 
 def is_jupyter() -> bool:
@@ -39,7 +35,7 @@ def is_databricks() -> bool:
         bool: True if the module is running on Databricks notebook, False otherwise.
     """
     # Note, this is a hacky way to check if the code is running on Databricks.
-    if str(Path(".").resolve()) == "/databricks/driver":
+    if "DATABRICKS_RUNTIME_VERSION" in os.environ:
         return True
     else:
         return False
--- a/feathr_project/feathr/utils/platform.py
+++ b/feathr_project/feathr/utils/platform.py
@@ -15,6 +15,10 @@ def is_jupyter() -> bool:
     Returns:
         bool: True if the module is running on Jupyter notebook or Jupyter console, False otherwise.
     """
+    # Should check is_databricks() and is_synapse() first since they also use the same ZMQ interactive shell.
+    if is_databricks() or is_synapse():
+        return False
+
     try:
         # Pre-loaded module `get_ipython()` tells you whether you are running inside IPython or not.
         shell_name = get_ipython().__class__.__name__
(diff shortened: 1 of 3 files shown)


xiaoyongzhu (dismissed):

feathr_project/setup.py
```diff
@@ -84,6 +84,7 @@
         "azure-storage-file-datalake<=12.5.0",
         "azure-synapse-spark<=0.7.0",
         # fixing Azure Machine Learning authentication issue per https://stackoverflow.com/a/72262694/3193073
+        "aiohttp==3.8.3",  # pin to resolve Synapse import issue
```
xiaoyongzhu, line 87: Can you elaborate a bit more on this? What is the specific issue?

loomlike: This is what I found from our synapse example notebook:

>If you meet errors like 'cannot import FeatherClient from feathr', it may be caused by incompatible version of 'aiohttp'. Please try to install/upgrade it by running: '%pip install aiohttp==3.8.3'

and that actually what I encountered when I tried the feathr package import on the synapse. I found pinning `aiohttp==3.8.3` resolves the issue.

loomlike: specifically, the message is `function() argument 'code' must be code, not str` throwing from aiohttp package that is installed on Synapse (version == 3.7.4.post0). I'll add this detail to the setup.py file.

feathr_project/feathr/utils/platform.py
```diff
@@ -33,13 +38,21 @@ def is_databricks() -> bool:
...
         bool: True if the module is running on Databricks notebook, False otherwise.
     """
-    try:
-        if str(Path(".").resolve()) == "/databricks/driver":
-            return True
-        else:
-            return False
-    except NameError:
+    # Note, this is a hacky way to check if the code is running on Databricks.
+    if str(Path(".").resolve()) == "/databricks/driver":
```
blrchen, line 42: Check DATABRICKS_RUNTIME_VERSION might be more stable

feathr_project/setup.py
```diff
@@ -83,6 +83,8 @@
         "avro<=1.11.1",
         "azure-storage-file-datalake<=12.5.0",
         "azure-synapse-spark<=0.7.0",
+        # Synapse's aiohttp package is old and does not work with Feathr. We pin to a newer version here.
+        "aiohttp==3.8.3",
```
blrchen, line 87: Suggest to clean up aiohttp notes in [product_recommendation_demo_advanced.ipynb](https://github.com/feathr-ai/feathr/blob/66480d7d9edbc0b8e4d7307e0fe6e5bad66a11f5/docs/samples/product_recommendation_demo_advanced.ipynb#L215) as user does not need to care about this any more.

loomlike: it's already done in my another PR refactoring the notebooks! Thank you for the comment.

feathr_project/feathr/utils/platform.py
loomlike: this is great. I just changed the codes. Thank you!

blrchen (approved):

xiaoyongzhu (approved):

PR #935, opened 2023-01-03
publish fat jar in maven update action

### LINKED ISSUE #934: [FR] Public Fat Jar should be published when maven jar is updated.
### Willingness to contribute

### Feature Request Proposal

Current `Publish to Maven` Github action only updates maven Jar. As we also provide a public fat jar runtime location to users, all released maven jar should have a related public fat jar to track and make sure the version aligned. 

### Motivation

the public fat jar: https://azurefeathrstorage.blob.core.windows.net/public/feathr-assembly-LATEST.jar needs to be updated, while there is no official released fat jar to copy from. 

### Details

_No response_

Yuqing-cat:
## Description

Resolves #934 

## How was this PR tested?

Currently, the Github Action will write fat jar to the storage defined by "SPARK_JAR_BLOB_CONNECTION_STRING". To separate it from CI artifacts, we may need to create a new secret to store released jar blob connection string in the future. 
This PR only provide a trackable released Fat jar for reference. 

### DIFF
--- a/.github/workflows/publish-to-maven.yml
+++ b/.github/workflows/publish-to-maven.yml
@@ -35,3 +35,21 @@ jobs:
           PGP_SECRET: ${{ secrets.MAVEN_GPG_PRIVATE_KEY }}
           SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
           SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
+
+    # Publish Released Fat Jar to Blob Storage 
+      - name: Gradle build
+        run: |
+          ./gradlew build
+          # remote folder for CI upload
+          echo "CI_SPARK_REMOTE_JAR_FOLDER=feathr_jar_release" >> $GITHUB_ENV
+          # get local jar name without path
+          echo "FEATHR_LOCAL_JAR_FULL_NAME_PATH=$(ls build/libs/*.jar)" >> $GITHUB_ENV
+      
+      - name: Azure Blob Storage Upload (Overwrite)
+        uses: fixpoint/azblob-upload-artifact@v4
+        with:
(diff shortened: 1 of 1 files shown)


xiaoyongzhu (approved):

.github/workflows/publish-to-maven.yml
```diff
@@ -35,3 +35,21 @@ jobs:
...
+          echo "CI_SPARK_REMOTE_JAR_FOLDER=feathr_jar_release" >> $GITHUB_ENV
+          # get local jar name without path
+          echo "FEATHR_LOCAL_JAR_FULL_NAME_PATH=$(ls build/libs/*.jar)" >> $GITHUB_ENV
+      
+      - name: Azure Blob Storage Upload (Overwrite)
+        uses: fixpoint/azblob-upload-artifact@v4
+        with:
+          connection-string: ${{secrets.SPARK_JAR_BLOB_CONNECTION_STRING}}
+          name: ${{ env.CI_SPARK_REMOTE_JAR_FOLDER}}
+          path: ${{ env.FEATHR_LOCAL_JAR_FULL_NAME_PATH}}
```
blrchen, line 53: Suggest change 

- `CI_SPARK_REMOTE_JAR_FOLDER` => `FEATHR_JAR_AZURE_STORAGE_FOLDER`
- `FEATHR_LOCAL_JAR_FULL_NAME_PATH` => `FEATHR_JAR_LOCAL_FULL_PATH`

Yuqing-cat: Those are the usage practice copied from https://github.com/feathr-ai/feathr/blob/main/.github/workflows/pull_request_push_test.yml, do you think I should update both?

blrchen: I didn't realize it's already in use, then let's avoid change names already in use.

blrchen (approved):

PR #954, opened 2023-01-10
Update registry start script to support `REACT_APP_ENABLE_RBAC==false` cases

### LINKED ISSUE #953: [BUG] `REACT_APP_ENABLE_RBAC=false` not recognized in registry service
### Willingness to contribute

### Feathr version

latest

### System information

- **Python version**:

### Describe the problem

REACT_APP_ENABLE_RBAC variable is checked either configured or not rather than `true` or `false`.
In provision template, it's null if not `true`, so will not be a problem.
```
"[if(equals(parameters('enableRBAC'),'Yes'),json('true'),json('null'))]"
```
However, It might cause problem when user manually set `REACT_APP_ENABLE_RBAC` == false. 

### Tracking information

_No response_

### Code to reproduce bug

_No response_

Yuqing-cat:
## Description

Resolves #953 

## How was this PR tested?

This PR only include a logic change in `start.sh`.
More info are described in issue #953

### DIFF
--- a/deploy/start.sh
+++ b/deploy/start.sh
@@ -33,19 +33,8 @@ nginx
 # Start API app
 LISTENING_PORT="8000"
 
-if [ "x$REACT_APP_ENABLE_RBAC" == "x" ]; then
-    echo "RBAC flag not configured, only launch registry app"
-    if [ "x$PURVIEW_NAME" == "x" ]; then
-        echo "Purview flag is not configured, run SQL registry"
-        cd sql-registry
-        uvicorn main:app --host 0.0.0.0 --port $LISTENING_PORT
-    else
-        echo "Purview flag is configured, run Purview registry"
-        cd purview-registry
-        uvicorn main:app --host 0.0.0.0 --port $LISTENING_PORT
-    fi
-else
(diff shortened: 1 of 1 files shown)


xiaoyongzhu (approved):

deploy/start.sh
```diff
@@ -33,19 +33,8 @@ nginx
...
-        cd sql-registry
-        uvicorn main:app --host 0.0.0.0 --port $LISTENING_PORT
-    else
-        echo "Purview flag is configured, run Purview registry"
-        cd purview-registry
-        uvicorn main:app --host 0.0.0.0 --port $LISTENING_PORT
-    fi
-else
-    echo "RBAC flag configured, launch both rbac and reigstry apps"
+if [ "$REACT_APP_ENABLE_RBAC" == "true" ]; then
```
blrchen, line 36: Can this work if env is set to True

Yuqing-cat: No, only works when it's set as "true", according to our provision code 
```
"[if(equals(parameters('enableRBAC'),'Yes'),json('true'),json('null'))]"
```

jainr (approved):

PR #965, opened 2023-01-13
Fix registry to support Source with types other than HDFS

windoze:
## Description

Fix registry to support Source with types other than HDFS.


registry/sql-registry/registry/models.py
```diff
@@ -674,16 +674,16 @@ def to_attr(self) -> ProjectAttributes:
...
                  qualified_name: str = "",
                  preprocessing: Optional[str] = None,
                  event_timestamp_column: Optional[str] = None,
                  timestamp_format: Optional[str] = None,
-                 tags: Dict = {}):
+                 tags: dict = {},
+                 **options: Dict[str, Any]):
         self.qualified_name = qualified_name
         self.name = name
-        self.path = path
```
xiaoyongzhu, line 686: Why is `path` removed?

windoze: It's now in options, and not limited to `path`.
The change is to support other source types, some of them don't have `path` attribute.

registry/sql-registry/registry/models.py
```diff
@@ -435,14 +435,14 @@ def __init__(self,
     def entity_type(self) -> EntityType:
         return EntityType.Source
 
-    def to_dict(self) -> Dict:
```
blrchen, line 438: Is it intented to change Dict to dict?

xiaoyongzhu (approved):

blrchen (approved):

PR #979, opened 2023-01-18
Updating automated maven publish workflow to work with gradle creds

jainr:
## Description
The gradle plugin being used to build and publish jar to maven central requires that credentials be passed a certain way.
As documented here- https://vanniktech.github.io/gradle-maven-publish-plugin/central/#secrets

## How was this PR tested?
Will Test it as part of PR workflow.

### DIFF
--- a/.github/workflows/publish-to-maven.yml
+++ b/.github/workflows/publish-to-maven.yml
@@ -27,17 +27,20 @@ jobs:
 
     # CI release command defaults to publishSigned
     # Sonatype release command defaults to sonaTypeBundleRelease
-      - name: Gradle publish
-        run: gradle clean publish
+    # Changing env names as documented here - https://vanniktech.github.io/gradle-maven-publish-plugin/central/#secrets
+      - name: Gradle Build and Publish to Maven central
+        run: |
+          ./gradlew publish
+          ./gradlew closeAndReleaseRepository
         env:
-          PGP_PASSPHRASE: ${{ secrets.MAVEN_GPG_PASSPHRASE }}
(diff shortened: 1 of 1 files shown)


rakeshkashyap123 (commented):
The changes look good to me. Can you comment on how we are planning to test it? Is it only upon checkin?

Yuqing-cat (approved):

.github/workflows/publish-to-maven.yml
```diff
@@ -27,17 +27,20 @@ jobs:
 
     # CI release command defaults to publishSigned
     # Sonatype release command defaults to sonaTypeBundleRelease
-      - name: Gradle publish
-        run: gradle clean publish
+    # Changing env names as documented here - https://vanniktech.github.io/gradle-maven-publish-plugin/central/#secrets
+      - name: Gradle Build and Publish to Maven central
+        run: |
+          ./gradlew publish
```
rakeshkashyap123, line 33: Also, do want to publish for all check-ins? or do we want to bring back the condition removed here?
https://github.com/feathr-ai/feathr/pull/974

jainr: This check makes sure it only gets triggered for a new tag

jainr: tags: ["*"]

jainr:
> The changes look good to me. Can you comment on how we are planning to test it? Is it only upon checkin?

Once the PR is checked in, delete one of the tags, recreate it and that should trigger a build.

rakeshkashyap123 (approved):

PR #987, opened 2023-01-18
Fix env logging

xiaoyongzhu:
This PR fixes a few environment variable related issues and tests. Specifically this PR removes the `use_env_vars` variable from all places since it probably doesn't make a lot of sense.


feathr_project/setup.py
```diff
@@ -65,6 +65,7 @@
         "py4j<=0.10.9.7",
         "loguru<=0.6.0",
         "pandas",
+        "numpy<=1.20.3",  # pin numpy due to pyspark's deprecated np.bool access
```
blrchen, line 68: Should this one get reverted?

xiaoyongzhu: I think so, but let's see if all the test can pass, then revert this one. Otherwise one of the test might fail

PR #988, opened 2023-01-18
Fix local submission issues.

xiaoyongzhu:
This PR solves the local spark submission issues by:
- adding new Maven sources
- Support using a pre-built jar in a Python fashion (PySpark)


feathr_project/feathr/spark_provider/_localspark_submission.py
```diff
@@ -106,7 +108,23 @@ def submit_feathr_job(
                 print(python_files)
                 spark_args.append(python_files[0])
         else:
-            spark_args.extend(["--class", main_class_name, main_jar_path])
+            if not python_files:
```
Yuqing-cat, line 111: If a main jar is provided, we can use this main jar directly rather than add any dependencies.

xiaoyongzhu: If a main jar is provided, we also need to consider if there's a python file or not. If there's a python file provided, we should still go thru the python path.

xiaoyongzhu: But I guess maybe we don't need to add those dependencies?

Yuqing-cat (approved):

blrchen (approved):

PR #992, opened 2023-01-18
Resolve pyspark / numpy conflicts

loomlike:

## Description
Pyspark is still relying on the old numpy api, referring `np.bool` that has been deprecated.
Because of that, when calling sparkDF.toPandas(), it throws `AttributeError: module 'numpy' has no attribute 'bool'`.
We have to upgrade pyspark version when they release new patch.


xiaoyongzhu (dismissed):

blrchen:
I am having concern to do version pin in Feathr. Its okay to do version pin in Registry as its a standalone app. But for feathr it's a library and normally used in an environment with other python libraries in same context. Introduce a version pin might introduces the risk for package installation error or incompatible issues with other python packages.

I actually already experienced some issue when numpy was pin earlier
- Nightly notebook test fails with error `ImportError: this version of pandas is incompatible with numpy < 1.20.3`
- Installation issue on python 3.10 `RuntimeWarning: NumPy 1.20.3 may not yet support Python 3.10`

And seems this is already fixed in Spark https://github.com/apache/spark/pull/37817, probably we can just set a min version for pyspeak instead?

image

image

loomlike:
@xiaoyongzhu @blrchen I verified that unless we explicitly call sparkDF.toPandas() on the dataframe that include boolean type features, we can avoid the pyspark's bug. Once pyspark has new release, let's change the pyspark dependency to `>new_version` to address this.

Until then, we can stick with the current version. I'll put the comment on our setup.py instead of pinning numpy.

blairch (approved):

xiaoyongzhu (approved):

PR #1008, opened 2023-01-21
Fix sandbox issues and improve experience

xiaoyongzhu:
This PR focuses on:

- Make the sandbox registry URL configurable so users can configure it using environment variable
- Make sure sandbox always use latest code on main
- Fix a few issues when Sandbox is using maven jars, sometimes it cannot find those jars.


feathr_project/feathr/spark_provider/_localspark_submission.py
```diff
@@ -205,6 +209,11 @@ def wait_for_completion(self, timeout_seconds: Optional[float] = 500) -> bool:
                 contents = f.read()
                 logger.error(contents)
             return False
+        elif proc.returncode == 143:
+            logger.info(
+                f"Spark job with pid {self.latest_spark_proc.pid} finished in: {int(job_duration)} seconds."
```
Yuqing-cat, line 214: Better include some hint about 143 returncode in log or add some comment here. 
AFAIK, when local spark terminate the job due to no response for certain seconds, it will return 143. However, there are still other cases that 143 is returned not by design.

registry/sql-registry/registry/db_registry.py
```diff
@@ -52,8 +52,10 @@ class DbRegistry(Registry):
     def __init__(self):
         self.conn = connect()
         if os.environ.get("FEATHR_SANDBOX"):
-            if os.environ.get("FEATHR_SANDBOX_REGISTRY_URL"):
-                engine = db.create_engine(f'sqlite:///{os.environ.get("FEATHR_SANDBOX_REGISTRY_URL")}?check_same_thread=False')
+            sandbox_registry_url = os.environ.get("FEATHR_SANDBOX_REGISTRY_URL")
```
Yuqing-cat, line 55: @enya-yx , based on our work on code coverage increase, will it be useful to add some "FEATHR_SANDBOX" specific test functions?

enya-yx: > @enya-yx , based on our work on code coverage increase, will it be useful to add some "FEATHR_SANDBOX" specific test functions?

Agree with that. To maintain the test coverage ratio and also provide some examples as a reference, I suggest to add some related test cases.

feathr_project/feathr/spark_provider/_localspark_submission.py
xiaoyongzhu: good point, added

registry/sql-registry/registry/db_registry.py
xiaoyongzhu: This is good point, unfortunatelly those code cannot be unit tested since it requires a docker env setup. 

For instructions, it's part of the readme file:

By default, Feathr Sandbox uses a SQLite backend in Feathr registry, and the content will be deleted if you restart the container. If you want to use a persistent registry, say a remote MySQL database or SQL Server, you can configure an environment variable when starting the Feathr Sandbox container like below, by specifying `FEATHR_SANDBOX_REGISTRY_URL` environment variable:

```bash
docker run -it --rm -p 8888:8888  -p 8000:8000 -p 8081:80 -p 8080:8080 -p 7080:7080 --env API_BASE="api/v1" --env FEATHR_SANDBOX=True --env FEATHR_SANDBOX_REGISTRY_URL="mysql://scott:tiger@localhost/foo" -e GRANT_SUDO=yes feathrfeaturestore/feathr-sandbox
```

Take a look at the [Database URLs](https://docs.sqlalchemy.org/en/20/core/engines.html#database-urls) section for more details on how this URL is formatted. Basically it will be like this:

Yuqing-cat (approved):

blrchen (approved):

PR #1011, opened 2023-01-25
Refactor config helper to be cleaner

loomlike:

## Description

Resolves #987

Also please do review `.github/workflows/pull_request_push_test.yml` file to make sure the changes are valid. I copied the file from #987.

## How was this PR tested?

Unittest

### DIFF
--- a/feathr_project/feathr/utils/_env_config_reader.py
+++ b/feathr_project/feathr/utils/_env_config_reader.py
@@ -49,20 +49,17 @@ class EnvConfigReader(object):
         Returns:
             Feathr client's config value.
         """
-        res_env = self._get_variable_from_env(key)
-        res_file = (self._get_variable_from_file(key) if self.yaml_config else None)
-        res_keyvault = (self._get_variable_from_akv(key) if self.akv_name else None)
+        val = self._get_variable_from_env(key)
+        if val is None and self.yaml_config is not None:
+            val = self._get_variable_from_file(key)
+        if val is None and self.akv_name is not None:
+            val = self._get_variable_from_akv(key)
 
-        # rewrite the logic below to make sure:
-        # First we have the order (i.e. res1 > res2 > res3 > default)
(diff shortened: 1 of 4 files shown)


feathr_project/feathr/utils/_env_config_reader.py
```diff
@@ -78,24 +75,20 @@ def get_from_env_or_akv(self, key: str) -> str:
...
-        # First we have the order (i.e. res1 > res2 > res3 > default)
-        # Also previously we use OR for the result, which will yield a bug where say res1=None, res2=False, res3=None. Using OR will result to None result, although res2 actually have value
-        for res in [res_env, res_keyvault]:
-            if res is not None:
-                return res
-
-        logger.warning(f"Config {key} is not found in the environment variable or the remote key value store.")
-        return None
+        if val is not None:
+            return val
```
xiaoyongzhu, line 83: @loomlike can you help to bring some of the previous comments to the updated code so folks are aware what's going on?

xiaoyongzhu (approved):

xiaoyongzhu:
Thanks @loomlike! This looks good to me. Can you also add some of the previous deleted comments to the updated code so folks are aware what's going on?

feathr_project/feathr/utils/_env_config_reader.py
xiaoyongzhu: Something like this:

```
# rewrite the logic below to make sure:
        # First we have the order (i.e. res1 > res2 > res3 > default)
        # Also previously we use OR for the result, which will yield a bug where say res1=None, res2=False, res3=None. Using OR will result to None result, although res2 actually have value
```

loomlike: The docstring of the function already contains that info (the order):

```
        It will retrieve the value in the following order:
            - From the environment variable if `use_env_vars == True` and the key is set in the os environment variables.
            - From the config yaml file if the key exists.
            - From the Azure Key Vault.
```
Let me know if you think this is clear enough.

xiaoyongzhu (approved):

jainr (approved):

PR #1023, opened 2023-01-31
Auto Delete Empty Anchors and Sources in SQL Registry

### LINKED ISSUE #1022: [FR] Empty Anchor and Source could be auto deleted.
### Willingness to contribute

### Feature Request Proposal

Current UI doesn't have "anchor" or "source" page to delete those entities, which blocks project deletion in UI.

### Motivation

Delete a whole project rather than several features.

### Details

_No response_

Yuqing-cat:
Signed-off-by: Yuqing Wei <weiyuqing021@outlook.com>

## Description

Resolves #1022 
Add a `delete_empty_entities ` function to clean up empty anchors and sources in given entity list. 

## How was this PR tested?

Project in SQL registry can be deleted successfully after all features are removed. 
image
image

### DIFF
--- a/registry/sql-registry/registry/db_registry.py
+++ b/registry/sql-registry/registry/db_registry.py
@@ -204,6 +204,28 @@ class DbRegistry(Registry):
             downstream_entities, _ = self._bfs(entity_id, RelationshipType.Produces)
         return [e for e in downstream_entities if str(e.id) != str(entity_id)]
     
+    def delete_empty_entities(self, entities: List[Entity]):
+        """
+        Given entity list, deleting all anchors that have no features and all sources that have no anchors.
+        """
+        if len(entities) == 0:
+            return
+        
+        # clean up empty anchors
+        for e in entities:
+            if e.entity_type == EntityType.Anchor:
+                downstream_entities, _ = self._bfs(e.id, RelationshipType.Contains) 
+                if len(downstream_entities) == 1: # only anchor itself
+                    self.delete_entity(e.id)
+        # clean up empty sources
+        for e in entities:
+            if e.entity_type == EntityType.Source:
+                downstream_entities, _ = self._bfs(e.id, RelationshipType.Produces) 
+                if len(downstream_entities) == 1: # only source itself
+                    self.delete_entity(e.id)
+
+        return
+        
     def delete_entity(self, entity_id: Union[str, UUID]):
         """
         Deletes given entity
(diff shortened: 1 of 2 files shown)


registry/sql-registry/registry/db_registry.py
```diff
@@ -204,6 +204,28 @@ def get_dependent_entities(self, entity_id: Union[str, UUID]) -> List[Entity]:
...
+        for e in entities:
+            if e.entity_type == EntityType.Anchor:
+                downstream_entities, _ = self._bfs(e.id, RelationshipType.Contains) 
+                if len(downstream_entities) == 1: # only anchor itself
+                    self.delete_entity(e.id)
+        # clean up empty sources
+        for e in entities:
+            if e.entity_type == EntityType.Source:
+                downstream_entities, _ = self._bfs(e.id, RelationshipType.Produces) 
+                if len(downstream_entities) == 1: # only source itself
```
xiaoyongzhu, line 224: Do we need to check the returned entity type? Just to make sure downstream_entities is Anchor/Source?

Yuqing-cat: the _bfs entity list return will always contain the entity itself (`e`), which type is already checked above (`e.entity_type`).

enya-yx: Is that better we use one for loop here? Eg. for e in entities: if type is Anchor: ...; elif type is Source: ...

Yuqing-cat: > Is that better we use one for loop here? Eg. for e in entities: if type is Anchor: ...; elif type is Source: ...

`Source` produces `Anchor`, they have sequence.  

If merged in one loop, for `Source`, you need to check if all the downstream entities has empty `Anchor` which also involves another loop.

But if all empty anchors are cleaned in the first loop, then we can clean empty sources directly.

xiaoyongzhu (approved):

registry/sql-registry/registry/db_registry.py
enya-yx: > > Is that better we use one for loop here? Eg. for e in entities: if type is Anchor: ...; elif type is Source: ...
> 
> `Source` produces `Anchor`, they have sequence.
> 
> If merged in one loop, for `Source`, you need to check if all the downstream entities has empty `Anchor` which also involves another loop.
> 
> But if all empty anchors are cleaned in the first loop, then we can clean empty sources directly.

Good point

enya-yx (approved):

PR #1027, opened 2023-02-01
Fix bug when SWA feature datapath does not end in daily/hourly

rakeshkashyap123:
When SWA does not have end in daily/hourly path, it fails. In this PR, we will add a corrective behavior to add the suffixes.


bozhonghu (approved):
LGTM

jaymo001 (approved):

PR #1028, opened 2023-02-02
Auto add resolution when timePartitionPattern is present

rakeshkashyap123:
When timePartitionPattern is present, the pathAnalyzer does not add the daily/hourly suffix, if not present.
In this PR, we fix that issue.

### DIFF
--- a/feathr-impl/src/main/scala/com/linkedin/feathr/offline/source/pathutil/TimeBasedHdfsPathAnalyzer.scala
+++ b/feathr-impl/src/main/scala/com/linkedin/feathr/offline/source/pathutil/TimeBasedHdfsPathAnalyzer.scala
@@ -11,6 +11,10 @@ import com.linkedin.feathr.offline.source.dataloader.DataLoaderHandler
  */
 private[offline] class TimeBasedHdfsPathAnalyzer(pathChecker: PathChecker, dataLoaderHandlers: List[DataLoaderHandler]) {
 
+  val dailyFolder = "daily/"
+  val hourlyFolder = "hourly/"
+  val dailyPattern = "yyyy/MM/dd"
+  val hourlyPattern = "yyyy/MM/dd/HH"
   /**
    * check whether the given path is daily or hourly partitioned.
    *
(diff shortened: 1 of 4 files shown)


bozhonghu (dismissed):
LGTM

feathr-impl/src/main/scala/com/linkedin/feathr/offline/source/pathutil/TimeBasedHdfsPathAnalyzer.scala
```diff
@@ -70,7 +70,18 @@ private[offline] class TimeBasedHdfsPathAnalyzer(pathChecker: PathChecker, dataL
...
+    } else if (pathChecker.exists(basePath + dailyFolder)) {
+      PathInfo(basePath + dailyFolder, dateTimeResolution, timePartitionPattern)
+    } else if (pathChecker.exists(basePath + hourlyFolder)) {
+      PathInfo(basePath + hourlyFolder, dateTimeResolution, timePartitionPattern)
+    } else {
+      // Daily data can be Orc/Hive data following in HomeDir/datepartition=yyyy-MM-dd-00
+      PathInfo(basePath + "datepartition=", dateTimeResolution, "yyyy-MM-dd-00")
+    }
   }
 }
```
enya-yx, line 86: Should we also concern below cases (not sure if they exist in real situations):
1. The base path doesn't end with daily/ or hourly/, the whole path is like: {somepath}/samples/2023/01/01/
2. The base path ends with daily/ but the time partition pattern is not /yyyy/MM/dd, the whole path is like: {somepath}/daily/2023-01-01

jainr:
Hi @rakeshkashyap123 , One of the gradle test is failing. Could you please take a look?
https://github.com/feathr-ai/feathr/actions/runs/4070502148/jobs/7011416482

feathr-impl/src/main/scala/com/linkedin/feathr/offline/source/pathutil/TimeBasedHdfsPathAnalyzer.scala
enya-yx: One of the gradle test case failed at 'TestTimeBasedHdfsPathAnalyzer.scala:78' (you can find more details from below link). Could you solve that before merging?

rakeshkashyap123: 1. These are mainly meant for hdfs paths. I dont think paths of that pattern exist, and if they do we may need to write another pathAnalyzer.
2. Fixed the test, thanks for pointing it out.

loomlike:
@rakeshkashyap123

There's error from the notebook tests. Found this from ADB job logs:

AnalysisException: Path does not exist: dbfs:/feathrazure_cijob_12_17_41374/nyc_taxi_data.csv/datepartition= 

Seems the code append `datapartition=` sufix to the path and that makes the path invalid.

enya-yx (approved):
LGTM

jainr (approved):
Thanks for fixing the tests, looks good.

PR #1043, opened 2023-02-08
Remove json schema validator

windoze:
## Description

- Remove JSON Schema Validator dependency so we won't need additional parameter to submit Spark job.
- Resolve the not found issue for non maven package in Azure Databricks/Synapse
- Publish a v0.10.4-rc5 to maven and test can pass.


blrchen (approved):

blrchen:
LGTM, the main reason of this removal is curent json validator code introduce a new dependency org.everit.json.schema;1.9.1 which is not available in maven central. This will results `org.everit.json.schema;1.9.1: not found` job runtime error when running on cloud sparks like Azure Databricks and Azure Synpase. This is a break change as user have to apply manual workaround on spark in order to make job succeed. 

Given the json schema validator is not used in current compute model, removal should be low risk if all CI test and notebook test can pass. We can revisit this when it's time to turn on vnext compute model, and explore more options to ensure critial cloud use case not break:
1. Downgrade to the old version org.everit.json.schema available on maven
2. Fork latest org.everit.json.schema code and submit to maven central 
3. Research how to automatically install non-maven packages on cloud sparks without manual workaround from end user side. 
4. Use a different json schema package which can be pulled from maven.

Yuqing-cat (approved):

PR #1050, opened 2023-02-09
Fix nyc_taxi data utils bug in databricks path

loomlike:

## Description
`nyc_taxi.get_spark_df` doesn't parses the `local_cache_path` argument correctly when it has "/dbfs/" prefix.
Previously, "/dbfs/myfile" is parsed as "dbfs:/dbfs/myfile", but it should be "dbfs:/myfile" instead.
This PR fixes this behavior.

## How was this PR tested?

Added more unit test cases after the fix.


blrchen (approved):

jainr (approved):
Looks good to me.

jainr:
After your PR, does this custom logic in sample notebook needs to be changed as well?
image

PR #1059, opened 2023-02-12
Add ability to extract features without any type conversion

rakeshkashyap123:
Add the ability to extract features as tensors only, without any type conversion.


jaymo001 (approved):

bozhonghu (approved):
LGTM

PR #1069, opened 2023-02-14
Debug and sanity mode

jaymo001:
## Description
1. Add sanity check mode. It will sample the source data and make sure key join matches. 
2. Add debug mode to provide more debugging information.
3. Refactor parameter setting code.


feathr-impl/src/main/scala/com/linkedin/feathr/offline/transformation/DataFrameExt.scala
```diff
@@ -38,5 +39,54 @@ private[offline] object DataFrameExt {
...
+     *
+     * @param leftJoinColumns
+     * @param rightJoinColumns
+     * @param rightDF
+     * @return
+     */
+    def appendRows(leftJoinColumns: Seq[String], rightJoinColumns: Seq[String], rightDF: DataFrame): DataFrame = {
+      // Add a few more rows to observation data by copying the feature key from the feature dataset
+      // This ensures the join will have matched results between observation and feature data
+      val dfCount = df.count()
```
rakeshkashyap123, line 67: Would'nt this be very expensive?

anirudhagar13:
One suggestion would be to have incremental smaller pull requests, which would be easier to review. It could be broken down into three parts, as described in the description of the pull request.

rakeshkashyap123 (approved):

feathr-impl/src/main/scala/com/linkedin/feathr/offline/transformation/DataFrameExt.scala
jaymo001: It should be fine as we are in sanity check mode where the dataframe typically has been down sampled to a few rows.

feathr-impl/src/main/scala/com/linkedin/feathr/offline/generation/SparkIOUtils.scala
```diff
@@ -74,8 +74,9 @@ object SparkIOUtils {
           val output_format = outputDF.sqlContext.getConf("spark.feathr.outputFormat", "avro")
           // if the output format is set by spark configurations "spark.feathr.outputFormat"
           // we will use that as the job output format; otherwise use avro as default for backward compatibility
-          outputDF.write.mode(SaveMode.Overwrite).format(output_format).save(path)
-          outputDF
+          if(!outputDF.isEmpty) {
```
anirudhagar13, line 77: Shouldn't we just write empty dataframe, instead of not producing an output?

anirudhagar13 (approved):
From a surface level, LGTM!

feathr-impl/src/main/scala/com/linkedin/feathr/offline/generation/DataFrameFeatureGenerator.scala
```diff
@@ -90,10 +87,13 @@ private[offline] class DataFrameFeatureGenerator(logicalPlan: MultiStageJoinPlan
...
           .map(f => (f._1, (offline.FeatureDataFrame(f._2.transformedResult.df, f._2.transformedResult.inferredFeatureTypes), f._2.joinKey)))
     }.toMap
+    allStageFeatures map { featureDfWithKey =>
+      FeathrUtils.dumpDebugInfo(ss, featureDfWithKey._2._1.df, Set(featureDfWithKey._1),
+      "transformed df in feature generation", "transformed_df_in_generation")
+    }
 
     // Update features based on skip missing feature flag and empty dataframe
-    val updatedAllStageFeatures = if (shouldSkipFeature || (ss.sparkContext.isLocal &&
-      SQLConf.get.getConf(LocalFeatureJoinJob.SKIP_MISSING_FEATURE))) {
```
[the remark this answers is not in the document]
jaymo001: This is a refactoring. The local config is now unified with production settings.

feathr-impl/src/main/scala/com/linkedin/feathr/offline/generation/SparkIOUtils.scala
jaymo001: It throws exception for empty DataFrame.

feathr-impl/src/main/scala/com/linkedin/feathr/offline/transformation/DataFrameExt.scala
```diff
@@ -38,5 +39,54 @@ private[offline] object DataFrameExt {
...
+     * 1  6  c  d
+     * 0  5  a  b
+     * 10  11  c d
+     *
+     * @param leftJoinColumns
+     * @param rightJoinColumns
+     * @param rightDF
+     * @return
+     */
+    def appendRows(leftJoinColumns: Seq[String], rightJoinColumns: Seq[String], rightDF: DataFrame): DataFrame = {
```
bozhonghu, line 64: Can we write a better documentation for this function. I don't quite understand what this function is supposed to do. One confusing part is the input is left join columns (seq of string), right join columns (seq of strings) and a rightDF (df), but the example in the docs seems to be showing two dataframes.

feathr-impl/src/main/scala/com/linkedin/feathr/offline/util/FeathrUtils.scala
```diff
@@ -128,4 +176,61 @@ private[offline] object FeathrUtils {
...
+    if (isDebugMode(ss)) {
+      val basePath = FeathrUtils.getFeathrJobParam(ss.sparkContext.getConf, FeathrUtils.DEBUG_OUTPUT_PATH)
+      val debugFeatureNames = FeathrUtils.getFeathrJobParam(ss.sparkContext.getConf, FeathrUtils.DEBUG_FEATURE_NAMES)
+        .split(FeathrUtils.STRING_PARAMETER_DELIMITER).toSet
+      val featureNames = "features_" + features.mkString("_") + "_"
+      if (debugFeatureNames.isEmpty || features.intersect(debugFeatureNames).nonEmpty) {
+        val savePath = SimplePath(basePath + "/" + featureNames + pathSuffix)
+        log.info(s"${tag}, Start dumping data ${featureNames} to ${savePath}")
+        if (!df.isEmpty) {
+          SparkIOUtils.writeDataFrame(df, savePath, Map(), List())
```
bozhonghu, line 203: With the large number of dumpDebugInfo calls, is there a concern that this will result in a really long run time and a huge file which is not really readable in the case of large dfs? Do we want to consider limiting the size of the df that gets written? If not then we should document the behavior of the debug mode clearly that the whole DF will get written to disk which can take long and be unreadable for large DFs.

bozhonghu (commented):
Couple comments regarding docs and concern regarding debug mode output.

feathr-impl/src/main/scala/com/linkedin/feathr/offline/util/FeathrUtils.scala
jaymo001: This is only for debug mode, the logging is turned off by default. We check the feature names in the dataframe so only the those specified in 'spark.feathr.debug.feature.names' will be dumped.

I tried to provide a filter before dump, however, the config would be non-trivial. It is essentially another config DSL, e.g. need to specify keys (may have multiple key columns) for each feature, observation at different steps. This is an overkill.

Dumping the some dataframe as a whole and using notebook to do analysis when needed is a better solution to me.

anirudhagar13 (approved):

PR #1093, opened 2023-02-27
Add support for external SWA library

rakeshkashyap123:
Add an optional support for handling SWA through other external libraries.


jaymo001 (approved):

feathr-impl/src/main/scala/com/linkedin/feathr/offline/swa/SlidingWindowAggregationJoiner.scala
```diff
@@ -25,6 +27,14 @@ import org.apache.spark.util.sketch.BloomFilter
 
 import scala.collection.mutable
 
+
+private[offline] case class SWAHandler(
+  join:
```
jaymo001, line 32: Nit: Better naming & documentation

jaymo001 (approved):

PR #1098, opened 2023-02-28
Implement an optional null filter before join

rakeshkashyap123:
In PR, add an optional flag to filter out the null key rows before a join. We will add back the null key rows after the join is complete. This helps in ensuring a faster join process.

This flag should be configured only when there are a lot of null keys to get performance bump, otherwise it can prove to be counter-productive.


jaymo001 (approved):

bozhonghu (approved):
LGTM. Would suggest we document every parameter like this for Feathr somewhere. Would be helpful in a wiki to know what are all the available options.

PR #1112, opened 2023-03-08
Support sql expression in FDSExtract

jaymo001:
## Description
Support sql expression in FDSExtract.


feathr-impl/src/main/scala/com/linkedin/feathr/offline/job/FeatureTransformation.scala
```diff
@@ -1359,7 +1359,7 @@ private[offline] object FeatureTransformation {
   def parseMultiDimTensorExpr(featureDef: String): String = {
     // String char should be one more than the len of the keyword to account for '('. The end should be 1 less than length of feature string
     // to account for ')'.
```
rakeshkashyap123, line 1361: Please update comment.

rakeshkashyap123 (approved):

anirudhagar13 (approved):
Not sure if it is allowed as it might be LI internal detail, but would be great to have some context as to what prompted the change, linked in the PR conversation.

PR #1127, opened 2023-03-29
Add a retryable mechanism for SWA features when files get deleted

rakeshkashyap123:
The retry will happen only once to ensure that the time to fail is not very high.

### DIFF
--- a/feathr-impl/src/main/scala/com/linkedin/feathr/offline/join/DataFrameFeatureJoiner.scala
+++ b/feathr-impl/src/main/scala/com/linkedin/feathr/offline/join/DataFrameFeatureJoiner.scala
@@ -327,7 +327,7 @@ private[offline] class DataFrameFeatureJoiner(logicalPlan: MultiStageJoinPlan, d
       offline.FeatureDataFrame(obsToJoinWithFeatures, Map())
     } else {
       val swaJoiner = new SlidingWindowAggregationJoiner(featureGroups.allWindowAggFeatures, anchorToDataSourceMapper)
-      swaJoiner.joinWindowAggFeaturesAsDF(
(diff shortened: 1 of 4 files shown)


feathr-impl/src/main/scala/com/linkedin/feathr/offline/swa/SlidingWindowAggregationJoiner.scala
```diff
@@ -256,8 +264,19 @@ private[offline] class SlidingWindowAggregationJoiner(
           }
         val origContextObsColumns = labelDataDef.dataSource.columns
 
-        contextDF = if (swaHandler.isDefined) swaHandler.get.join(labelDataDef, factDataDefs.toList) else SlidingWindowJoin.join(labelDataDef, factDataDefs.toList)
+        try {
+          // THIS IS FOR LOCAL TEST ONLY. It is to induce a spark exception with the root cause of FileNotFoundException.
+          if (isRetry && LocalFeatureJoinJob.shouldRetryAddingSWAFeatures) throw new SparkException("file not found", new FileNotFoundException())
```
jaymo001, line 269: Avoid using LocalFeatureJoinJob in non-testing code would be better., e.g. follow the con.spark.feathr.xx convention.

feathr-impl/src/main/scala/com/linkedin/feathr/offline/join/DataFrameFeatureJoiner.scala
```diff
@@ -338,6 +338,25 @@ private[offline] class DataFrameFeatureJoiner(logicalPlan: MultiStageJoinPlan, d
         swaObsTime,
         failOnMissingPartition,
         swaHandler)
+
+      // We will retry the SWA features which could not added because of
```
jaymo001, line 342: Fix comment

PR #1130, opened 2023-04-03
Fix bug when SWA hdfs and local paths without data.avro.json extensio…

rakeshkashyap123:
-> In current SWA, we filter out all the folders daily/hourly file paths which do not exist.
     For example, if a window of 7d is set, we look back at the past 7 days and filter out the paths which do not exist.
-> In this PR, we are making an additional change to ensure that the paths also non-empty. The directory of the folder should have some data in it, otherwise it would just at the time of loading the dataframe.
-> A lot of bugs are related to not being able to create a dataframe from a given path. This is because we do not filter out the empty folders while generating the paths.

### DIFF
--- a/feathr-impl/src/main/scala/com/linkedin/feathr/offline/source/pathutil/LocalPathChecker.scala
+++ b/feathr-impl/src/main/scala/com/linkedin/feathr/offline/source/pathutil/LocalPathChecker.scala
@@ -11,7 +11,7 @@ import java.io.File
  * @param hadoopConf hadoop configuration
  */
 private[offline] class LocalPathChecker(hadoopConf: Configuration, dataLoaderHandlers: List[DataLoaderHandler]) extends PathChecker {
-  val TEST_AVRO_JSON_FILE = "/data.avro.json"
+  private val TEST_AVRO_JSON_FILE = "/data.avro.json"
 
   /**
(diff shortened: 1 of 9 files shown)


minhmo1620 (dismissed):
LGTM

feathr-impl/src/main/scala/com/linkedin/feathr/offline/source/pathutil/LocalPathChecker.scala
```diff
@@ -53,4 +54,25 @@ private[offline] class LocalPathChecker(hadoopConf: Configuration, dataLoaderHan
...
+
+  /**
+   * Checks if the given path is non empty. If the path is a directory, check if it has files listed under it. Else, just check the
+   * file length.
+   * @param path input path
+   * @return true if the path is non empty.
+   */
+  override def nonEmpty(path: String): Boolean = {
+    if (!isExternalDataSource(path) && HdfsUtils.nonEmpty(path)) return true
+    val filePath = if (LocalFeatureJoinUtils.getMockPathIfExist(path, hadoopConf, None).isDefined) {
```
anirudhagar13, line 66: Nit: Could use isMock() in the same file.

anirudhagar13 (approved):
LGTM

PR #1138, opened 2023-04-05
Remove print statements

rakeshkashyap123:
Remove random print statements as it causes the node's memory to blow up.

### DIFF
--- a/feathr-impl/src/main/scala/com/linkedin/feathr/offline/job/FeatureTransformation.scala
+++ b/feathr-impl/src/main/scala/com/linkedin/feathr/offline/job/FeatureTransformation.scala
@@ -888,7 +888,6 @@ private[offline] object FeatureTransformation {
     val features = transformers map {
       case extractor: AnchorExtractor[IndexedRecord] =>
         val features = extractor.getFeatures(record)
-        print(features)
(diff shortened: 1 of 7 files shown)


anirudhagar13 (commented):
Is the goal to remove all print statements from feathr or the ones that output a lot of information?

If it is the former, then there are other print statements e.g.
print("in SQLConfigurableAnchorExtractor = " + newExtractor) in AnchorUDFOperator. and a couple of other files.

feathr-impl/src/main/scala/com/linkedin/feathr/offline/job/FeatureTransformation.scala
```diff
@@ -888,7 +888,6 @@ private[offline] object FeatureTransformation {
     val features = transformers map {
       case extractor: AnchorExtractor[IndexedRecord] =>
         val features = extractor.getFeatures(record)
-        print(features)
```
otterc, line 891: There is a `print(features)` at line 1426 as well

rakeshkashyap123: Removed.

PR #1146, opened 2023-04-10
#latest should pick the latest available path

rakeshkashyap123:
When users set their feature datasets to the #LATEST, the latest available folder should be picked up. Multiple times, we pick out the folder where is not mature, and this leads to errors.


minhmo1620 (dismissed):
Thanks for your fix!

feathr-impl/src/test/scala/com/linkedin/feathr/offline/source/accessor/TestDataSourceAccessor.scala
```diff
@@ -81,4 +81,15 @@ class TestDataSourceAccessor extends TestFeathr {
     assertTrue(accessor.isInstanceOf[NonTimeBasedDataSourceAccessor])
     assertEquals(accessor.get().count(), 10L)
   }
+
+  @Test(description = "test loading dataframe with BatchDataLoader having #LATEST in its path, also ensures that empty folders are not picked out.")
+  def testBatchDataLoaderWithLatestPathAndEmptyFolders(): Unit = {
+    // avroLatest folder has an empty folder with 10/01.
+    val path = "src/test/resources/avroLatest/#LATEST/#LATEST/#LATEST"
```
anirudhagar13, line 88: Apologies, I could not find an avroLatest folder with an empty folder here: https://github.com/feathr-ai/feathr/tree/main/feathr-impl/src/test/resources

Is it being added as a part of this commit? In this commit as well I only see 2022/09/15. Maybe empty folders are being removed by .gitignore

rakeshkashyap123: Good catch. Apparently, git does not include empty folders and we need to add a .gitkeep file in it. Added that.

anirudhagar13 (approved):

PR #1149, opened 2023-04-21
WIP: safe mode

rakeshkashyap123:
## Description

Resolves #XXX

### DIFF
--- a/feathr-impl/src/main/scala/com/linkedin/feathr/offline/swa/SlidingWindowAggregationJoiner.scala
+++ b/feathr-impl/src/main/scala/com/linkedin/feathr/offline/swa/SlidingWindowAggregationJoiner.scala
@@ -99,6 +99,8 @@ private[offline] class SlidingWindowAggregationJoiner(
     val timeWindowJoinSettings = joinConfigSettings.get.joinTimeSetting.get
     val simulatedDelay = timeWindowJoinSettings.simulateTimeDelay
     val shouldFilterNulls = FeathrUtils.getFeathrJobParam(ss.sparkContext.getConf, FeathrUtils.FILTER_NULLS).toBoolean
+    val shouldSkipFeature = FeathrUtils.getFeathrJobParam(ss.sparkContext.getConf, FeathrUtils.SKIP_MISSING_FEATURE).toBoolean
+    val isSafeMode = FeathrUtils.getFeathrJobParam(ss.sparkContext.getConf, FeathrUtils.SAFE_MODE).toBoolean
 
     if (simulatedDelay.isEmpty && !joinConfig.featuresToTimeDelayMap.isEmpty) {
       throw new FeathrConfigException(
@@ -137,6 +139,9 @@ private[offline] class SlidingWindowAggregationJoiner(
     // If the skip_missing_features flag is set, we will skip joining the features whose is not mature, and maintain this list.
     val notJoinedFeatures = new mutable.HashSet[String]()
 
+    // If the safe mode is turned off, we will add a null feature column and substitute it with the defaults if present.
+    val emptyFeatures = new mutable.HashSet[String]()
+
     // For each source, we calculate the maximum window duration that needs to be loaded across all
     // required SWA features defined on this source.
     // Then we load the source only once.
@@ -164,8 +169,11 @@ private[offline] class SlidingWindowAggregationJoiner(
               failOnMissingPartition)
 
         // If skip missing features flag is set and there is a data related error, an empty dataframe will be returned.
-        if (originalSourceDf.isEmpty) {
+        if (originalSourceDf.isEmpty && shouldSkipFeature) {
           res.map(notJoinedFeatures.add)
+          anchors.map(anchor => (anchor, originalSourceDf))
+        } else if (originalSourceDf.isEmpty && !isSafeMode) {
+          res.map(emptyFeatures.add)
           anchors.map(anchor => (anchor, originalSourceDf))
         } else {
         val sourceDF: DataFrame = preprocessedDf match {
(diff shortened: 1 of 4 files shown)


feathr-impl/src/main/scala/com/linkedin/feathr/offline/client/FeathrClient.scala
```diff
@@ -68,6 +68,15 @@ class FeathrClient private[offline] (sparkSession: SparkSession, featureGroups:
     SparkFeaturizedDataset(joinedDF, FeaturizedDatasetMetadata(header=Some(header)))
   }
 
+  /**
+   * APU which joins features and also returns any swallowed exception msgs.
+   */
+  @InternalApi
+  def joinFeaturesWithSwallowedExceptions(joinConfig: FeatureJoinConfig, obsData: SparkFeaturizedDataset,
+    jobContext: JoinJobContext = JoinJobContext()): (SparkFeaturizedDataset, String) = {
```
anirudhagar13, line 76: I think it would be better to have a return type to be more extensible, like a Map. That would make the API more generic. 

This would be to address any other future use cases where we want to send some other data back to FFE from the internal feathr engine. 

Therefore, we could have an API called `joinFeaturesExtended` OR `joinFeaturesV2` OR `joinFeaturesWithDataReturn`. 

LMK what you think. The issue with this design is that we would have to change it if we decide to send some other data back from the engine.

feathr-impl/src/main/scala/com/linkedin/feathr/offline/job/LocalFeatureJoinJob.scala
```diff
@@ -52,7 +52,7 @@ object LocalFeatureJoinJob {
       outputPath)
 
     val jobContext = FeatureJoinJob.parseInputArgument(defaultParams ++ extraParams).jobJoinContext
-    SparkFeaturizedDataset(feathrClient.joinFeatures(joinConfig, observationData, jobContext).data, FeaturizedDatasetMetadata())
+    SparkFeaturizedDataset(feathrClient.joinFeaturesWithSwallowedExceptions(joinConfig, observationData, jobContext)._1.data, FeaturizedDatasetMetadata())
```
anirudhagar13, line 55: Could we add a comment here describing the context of why we are accessing a specific index.

feathr-impl/src/main/scala/com/linkedin/feathr/offline/client/FeathrClient.scala
```diff
@@ -68,6 +68,15 @@ class FeathrClient private[offline] (sparkSession: SparkSession, featureGroups:
...
+    (joinFeatures(joinConfig, obsData, jobContext), SwallowedExceptionHandlerUtils.swallowedExceptionMsgs)
```
anirudhagar13, line 77: I think SwallowedExceptions seems a bit ambiguous. Could we change it something like `MutedExceptions` or `SupressedExceptions`?

feathr-impl/src/main/scala/com/linkedin/feathr/offline/swa/SlidingWindowAggregationJoiner.scala
```diff
@@ -164,9 +171,16 @@ private[offline] class SlidingWindowAggregationJoiner(
...
 
         // If skip missing features flag is set and there is a data related error, an empty dataframe will be returned.
-        if (originalSourceDf.isEmpty) {
+        if (originalSourceDf.isEmpty && shouldSkipFeature) {
           res.map(notJoinedFeatures.add)
           anchors.map(anchor => (anchor, originalSourceDf))
+        } else if (originalSourceDf.isEmpty && shouldAddDefaultColForMissingData) { // If add default col for missing data flag features
+          // flag is set and there is a data related error, an empty dataframe will be returned.
+          res.map(emptyFeatures.add)
+          val exceptionMsg = s"Missing data for features ${emptyFeatures}. Default values will be populated for this column."
```
anirudhagar13, line 180: The notification message itself would be carved in FFE. I think feathr-engine should just send back a Map of features patched with values substituted (either default or null when default is not defined).

feathr-impl/src/main/scala/com/linkedin/feathr/offline/swa/SlidingWindowAggregationJoiner.scala
```diff
@@ -137,6 +140,10 @@ private[offline] class SlidingWindowAggregationJoiner(
     // If the skip_missing_features flag is set, we will skip joining the features whose is not mature, and maintain this list.
     val notJoinedFeatures = new mutable.HashSet[String]()
 
+    // If the add default col for missing data flag is turned off,
```
anirudhagar13, line 143: Nit: could we edit the comment to have add_default_col_for_missing_data as one word. I got confused on reading the first time.

feathr-impl/src/test/scala/com/linkedin/feathr/offline/SlidingWindowAggIntegTest.scala
```diff
@@ -634,6 +633,98 @@ class SlidingWindowAggIntegTest extends FeathrIntegTest {
...
+        |        type: NUMERIC
+        |     }
+        |    }
+        |  }
+        |}
+        |derivations: {
+        | derived_simpleFeature: simpleFeature
+        |}
+      """.stripMargin
+    val res = runLocalFeatureJoinForTest(joinConfigAsString, featureDefAsString, observationDataPath = "slidingWindowAgg/localAnchorTestObsData.avro.json").data
```
anirudhagar13, line 720: Should we also add a unit test for API, i.e. if it returns back the proper Map of features patched?

feathr-impl/src/main/scala/com/linkedin/feathr/offline/client/FeathrClient.scala
anirudhagar13: An example of a metatdata map expected to return:
```
{
"features-with-missing-data": <feature-name-to-value-patched Map (to be sent over Iris notifications)>
"features-with-missing-data-exception": <complete exception string (to be logged in Azkaban logs)>
}
```

feathr-impl/src/main/scala/com/linkedin/feathr/offline/client/FeathrClient.scala
rakeshkashyap123: Good suggestion, changed it to SuppressedExceptions.

feathr-impl/src/main/scala/com/linkedin/feathr/offline/swa/SlidingWindowAggregationJoiner.scala
rakeshkashyap123: Done.

feathr-impl/src/main/scala/com/linkedin/feathr/offline/client/FeathrClient.scala
rakeshkashyap123: Done.

feathr-impl/src/test/scala/com/linkedin/feathr/offline/SlidingWindowAggIntegTest.scala
rakeshkashyap123: Done.

anirudhagar13 (approved):
Thank you for the fix!

PR #1152, opened 2023-04-26
Suppressed exceptions api

rakeshkashyap123:
Add another API which calls doJoinObsAndFeatures with suppressed exceptions.

### DIFF
--- a/feathr-impl/src/main/scala/com/linkedin/feathr/offline/client/FeathrClient.scala
+++ b/feathr-impl/src/main/scala/com/linkedin/feathr/offline/client/FeathrClient.scala
@@ -84,6 +84,22 @@ class FeathrClient private[offline] (sparkSession: SparkSession, featureGroups: 
       -> SuppressedExceptionHandlerUtils.missingDataSuppressedExceptionMsgs))
   }
 
+
+  /**
+   * API which joins features and also returns any suppressed exception msgs.
+   *
+   * @param joinConfig feathr offline's [[FeatureJoinConfig]]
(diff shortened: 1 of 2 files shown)


feathr-impl/src/main/scala/com/linkedin/feathr/offline/client/FeathrClient.scala
```diff
@@ -84,6 +84,22 @@ class FeathrClient private[offline] (sparkSession: SparkSession, featureGroups:
...
+   * @param joinConfig feathr offline's [[FeatureJoinConfig]]
+   * @param obsData    Observation data taken in as a [[SparkFeaturizedDataset]].
+   * @param jobContext [[JoinJobContext]]
+   * @return Joined observation and feature data as a SparkFeaturizedDataset and a map of suppressed exceptions along with
+   *         type of exception.
+   */
+  def doJoinObsAndFeaturesWithSuppressedExceptions(joinConfig: FeatureJoinConfig, obsData: SparkFeaturizedDataset,
+    jobContext: JoinJobContext = JoinJobContext()): (SparkFeaturizedDataset, Map[String, String]) = {
+    (joinFeatures(joinConfig, obsData, jobContext), Map(SuppressedExceptionHandlerUtils.MISSING_DATA_EXCEPTION
+      -> SuppressedExceptionHandlerUtils.missingDataSuppressedExceptionMsgs))
```
anirudhagar13, line 100: Could we remove specific date types here, as that is defeating the purpose of having a generic Map? Instead have string -> string datatype

anirudhagar13 (approved):

PR #1153, opened 2023-04-27
Fix doJoinObsAndFeaturesWithSuppressedExceptions API

rakeshkashyap123:
The API should call doJoinObsAndFeatures method, instead of the joinFeatures API.


feathr-impl/src/main/scala/com/linkedin/feathr/offline/client/FeathrClient.scala
```diff
@@ -95,8 +95,9 @@ class FeathrClient private[offline] (sparkSession: SparkSession, featureGroups:
    *         type of exception.
    */
   def doJoinObsAndFeaturesWithSuppressedExceptions(joinConfig: FeatureJoinConfig, obsData: SparkFeaturizedDataset,
```
anirudhagar13, line 97: Missing annotation for @InternalApi and we don't need the function just about this now right?

anirudhagar13 (approved):

PR #1161, opened 2023-05-08
Fix bug when skipping missing feature data

rakeshkashyap123:
Bug:-

When skipping features, the innermost method was not identifying the shouldAddDefaultCol flag.


anirudhagar13 (approved):
Thanks for this fix.

PR #1164, opened 2023-05-10
Fix bug when skipping anchored features with missing data

rakeshkashyap123:
The functionality would not work when more than one feature is present. We fix that bug in this PR.


anirudhagar13 (approved):
Thanks for the fix.

PR #1171, opened 2023-05-17
Fix bug in SWA with missing feature data

rakeshkashyap123:
1. While default values for missing SWA features, we do not care for the SWA groupings and add all the values together. This causes one of the asserts to fail.
2. Fix issue with mkString.


anirudhagar13 (approved):
Thanks for the fix!

PR #1196, opened 2023-06-27
Add String Column Support for Count Distinct Aggregation

atangwbd:
## Add String Support for Count Distinct Aggregation
Currently, feathr doesn't support the COUNT_DISTINCT aggregation on string column types, since it assumes the data type from the schema prior to the aggregation. When we try applying COUNT_DISTINCT on string columns, we get this error: 

`Caused by: Job aborted due to stage failure: Error while encoding: java.lang.RuntimeException: java.lang.Integer is not a valid external type for schema of string`

This PR converts each string into a unique 32 bit number using MD5 hash algorithm such that COUNT_DISTINCT aggregations can also work on string columns. 

Resolves #XXX

## How was this PR tested?
I ran spark jobs locally and saw them fail before this change, and succeed after this change.

COUNT_DISTINCT should now work on string columns.

- [x ] Yes. Make sure to clarify your proposed changes.


aabbasi-hbo (dismissed):

atangwbd:
> @atangwbd looks like there's a test failure? do you mind fixing it?
> 
> Gradle suite > Gradle test > com.linkedin.feathr.offline.SlidingWindowAggIntegTest > testSWACountDistinct FAILED org.apache.spark.sql.AnalysisException at SlidingWindowAggIntegTest.scala:1823

Fixed.

PR #1219, opened 2023-09-09
[Hdfs Utils] add uri param to filesystem get api

bheroder:
## Description

when HDFS is configured with fs.defaultsFS pointing to blob storage, we are not able to give local path as source location in feature definition config. this is because hdfsutils.scala uses FileSystem.get(conf) API. instead using FileSystem.get(URI, conf) solves the problem where one can give source location as file:///xyz even though defaultFS points to blob.

## Testing
manually tested using notebook session configured with fs.defaultFS= wasbs://xyz
FileSystem.get(conf).exists("file:///path") returns false
FileSystem.get(URI, conf).exists("file:///path") returns true

- [X ] No. You can skip the rest of this section.


bheroder:
cc @xiaoyongzhu

bheroder:
@rakeshkashyap123 could you help with tagging the right reviewers ?

bheroder:
@jaymo001 @rakeshkashyap123 could you please review ? cc @windoze

